PR #67, opened 2014-01-23
Add functions -setq and -let.

rejeep:
Hey,

Sometimes I work with key/value lists and I have to do:

```lisp
(let ((list
       '((foo "0.0.1")
         (bar "0.0.2")
         (baz "0.0.3")
         (qux "0.0.4"))))
  (-each list
    (lambda (item)
      (let ((name (car item))
            (version (cadr item)))
        (print name)
        (print version)))))
```

With this change, I can now do:

```lisp
(let ((list
       '((foo "0.0.1")
         (bar "0.0.2")
         (baz "0.0.3")
         (qux "0.0.4"))))
  (-each list
    (lambda (item)
      (-let (((name version) item))
        (print name)
        (print version)))))
```

This could be further extended with "pattern matching" support, like this:

```lisp
(-let (((_ &rest matches) (s-match "\\(.+\\)\\.\\(.+\\)" "foo.bar")))
  (print matches))
```

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,8 @@ Include this in your emacs settings to get syntax highlighting:
 * [-when-let*](#-when-let-vars-vals-rest-body) `(vars-vals &rest body)`
 * [-if-let](#-if-let-var-val-then-rest-else) `(var-val then &rest else)`
 * [-if-let*](#-if-let-vars-vals-then-rest-else) `(vars-vals then &rest else)`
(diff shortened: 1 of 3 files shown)


Fuco1:
For the missing `-setq` feature, see #116.

Fuco1:
FYI there is now https://github.com/magnars/dash.el/pull/271

PR #101, opened 2014-11-02
Update dash.el

ThibautVerron:
Hello,

I notice some big performance differences in calls of `-union` on my (rather slow) computer: computing the union of an empty list and a list with ~5000 elements takes about 1.5s, while computing the union of the same ~5000-long list and an empty list takes 0.02s.

The reason for that is clear by the implementation, and I don't think it should be changed. But I suggest a change in the docstring to make users aware of this phenomenon.

Note that other functions may show the same behavior, I only use a small subset of the library so I wouldn't know.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1561,7 +1561,12 @@ Alias: `-uniq'"
 (defun -union (list list2)
   "Return a new list containing the elements of LIST1 and elements of LIST2 that are not in LIST1.
 The test for equality is done with `equal',
-or with `-compare-fn' if that's non-nil."
+or with `-compare-fn' if that's non-nil.
+
+In case one of the list is much shorter than the other, the performances will be better if `-union` 
+is called with first argument the long list, like:
+
+  (-union long-list short-list)"
   (let (result)
     (--each list (!cons it result))
     (--each list2 (unless (-contains? result it) (!cons it result)))


Fuco1:
@doublep What do you think about this issue? Do you have some clever idea of optimizing it a bit?

doublep:
Well, natural approach would be to use a hash table to test for presence, since we may do that several times, depending on length of `list2`. I guess the best would be first compute length, if it is just a few items, use current implementation, else use a hash table.

Fuco1:
I speculated a while ago with sorting and doing the one-pass merge/intersection/deduplication but the problem is that the data can be unsortable.  This would lead to nlogn algorithm.

Is it always possible to take hash of any random value? (I'm not quite sure how emacs does that, I can't remember I ever used a hash table in elisp :D).  Hash table also eats a lot of memory, but I think it's a reasonable trade-off if the lists are big.

doublep:
As far as I understand it is possible, but it is not possible to create a hash table with arbitrary comparator. You need to define hash computer for each special comparator (i.e. equality function) you want to use. However, for `eq`, `equal` and `eql` hash computers are predefined, and these are the most important cases. If a custom comparator function is supplied we'd have to fallback on current slow implementation.

Fuco1:
Okey, I'll play with this tomorrow on the train home :) Thanks for the input.

PR #112, opened 2015-01-12
Add --mapc and for symmetry -mapc

tarsius:
I often wish I could use `--map` instead of `dolist` or `mapc` but when I do then the byte-compiler rightfully complains that I shouldn't be using `mapcar` when throwing away the value. So here's `--mapc` and for symmetry `-mapc` too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,7 @@ Convenient versions of `let` and `let*` constructs combined with flow control.
 
 Functions iterating over lists for side-effect only.
 
+* [-mapc](#-mapc-fn-list) `(fn list)`
 * [-each](#-each-list-fn) `(list fn)`
 * [-each-while](#-each-while-list-pred-fn) `(list pred fn)`
 * [-dotimes](#-dotimes-num-fn) `(num fn)`
@@ -1854,9 +1855,23 @@ See [`-let`](#-let-varlist-rest-body) for the description of destructuring mecha
 
 Functions iterating over lists for side-effect only.
 
+#### -mapc `(fn list)`
+
+Apply `fn` to each element of `list` for side effects only.
+
+```el
+(-mapc 'insert '("Hello " "side-effect\n"))
+(--mapc (insert "Hello " it "\n") '("Magnar" "Matus"))
+```
+
 #### -each `(list fn)`
 
 Call `fn` with every item in `list`. Return nil, used for side-effects only.
+
+This is similar to `-mapc` but the order of the arguments is reversed
+and the additional variable `it-index` is bound to integers from 0 to
+the length of the `list` minus one.  For performance criticial tasks
+`-mapc` should be used since it is more efficient.
 
 ```el
 (let (s) (-each '(1 2 3) (lambda (item) (setq s (cons item s))))) ;; => nil
--- a/dash.el
+++ b/dash.el
@@ -119,6 +119,15 @@ Return nil, used for side-effects only."
   (declare (debug (form form)))
   `(mapcar (lambda (it) ,form) ,list))
 
+(defun -mapc (fn list)
+  "Apply FN to each element of LIST for side effects only."
+  (mapc fn list))
+
+(defmacro --mapc (form list)
+  "Anaphoric form of `-mapc'."
+  (declare (debug (form form)))
+  `(mapc (lambda (it) ,form) ,list))
+
 (defmacro --reduce-from (form initial-value list)
   "Anaphoric form of `-reduce-from'."
   (declare (debug (form form form)))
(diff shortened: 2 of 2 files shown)


Fuco1:
Dash has `--each` which is like `mapc` with some extra features: like providing `it-index`.  I think better solution then would be to alias `-mapc` to `-each`?

Fuco1:
Granted, `mapc` is faster than `-each` because it's C while `-each` is built on top of a while loop.  Hmm... Do you think it is worth to optimize here?

tarsius:
I don't really intend to use `-mapc`, I'll just use `mapc` or `-each` as appropriate. So the performance difference does not really matter to me, but I tend to think this should not be an alias.

Making `-mapc` an alias for `-each` complicates documentation. Currently we have e.g.:

```
#### -map-when `(pred rep list)`
...
Alias: `-replace-where`
```

Meaning `-replace-where` is an alias for `-map-when` **and** `--replace-where` is an alias for `--map-when`. But while `-mapc` could be an alias for `-each`, `--mapc` isn't an alias for `--each`, so things would start falling apart documentation wise.

tarsius:
Or are you saying that not even `--mapc` is necessarily useful because there's already `--each`? In that case, I would argue that the better performance of  `--mapc` would indeed be desirable and justify its addition.

Fuco1:
That's what I ment yes, to alias both `-mapc` and `--mapc` to respective versions of `each`.

tarsius:
Actually that's not gonna work because the order of the arguments is different.

(I just pushed an amended commit, which puts the documentation in the right place).

Fuco1:
Oh, I see, you're right.

I guess each was mostly ment to be used in the `--each` variant in a way similar to (fictional) `with-list-items`, that is, the first argument declared special and the rest forms magically populated with data (cf. `with-current-buffer`, `with-temp-file` etc.)

Fuco1:
I think I'm okey with merging this, let's wait a couple days if @magnars shows up.

But you should add some description to the docstring as to why/how this is different to `-each` since they now duplicate behaviour and differ only in implementation (`-mapc` being somewhat faster as it is C basd).  Also a "see also" link could be added.

In other news, this patch breaks some compilation tests... and since compilation in emacs is something I totally ignore, I don't know how to fix it.

magnars:
How is it different than `-each`? Maybe it could be an alias? And if this one is faster, maybe `-each` could get a speed improvement?

fbergroth:
Regarding the failed test; the old test was broken and #113 hopefully fixes it.

Fuco1:
@magnars `mapc` iteration is implemented in C, whereas `-each` uses a while loop. That allows for some additional features like `it-index` which are otherwise (without using lexical scoping) impossible to achieve using just `mapc` and function hacking. They also have swapped arguments.

I don't know how much faster it is though, it would be cool to have some benchmarks.

fbergroth:
If it's just a matter of performance, we could scan `body` argument in `--each` for occurrences of `it-index`, and generate code using `mapc` if none are found. Also `it-index` seems to be undocumented in `--each`.

magnars:
Sounds like there is room for both. Would you mind rebasing on top of master, @tarsius - to run the tests again with the fixes?

tarsius:
I have rebased and updated the documentation, and have pushed the result.

When I run `./run-tests.sh` as-is I get:

```
dev/ert.el: `flet' is an obsolete macro (as of 24.3); use either `cl-flet' or `cl-letf'.
dev/ert.el: `labels' is an obsolete macro (as of 24.3); use `cl-labels' instead.
dev/ert.el: `flet' is an obsolete macro (as of 24.3); use either `cl-flet' or `cl-letf'.
Eager macro-expansion failure: (wrong-type-argument . "match-form must be a list")
Running 131 tests (2015-01-13 18:38:53+0100)
   passed    1/131  !cdr
...
   passed  131/131  -zip-with

Ran 131 tests, 131 results as expected (2015-01-13 18:38:54+0100)

Wrote /home/jonas/.emacs.d/lib/dash/dash.elc
```

When I remove `-l dev/ert.el` which shouldn't be required on `24.4` it get's worse:

```
Eager macro-expansion failure: (wrong-type-argument . "match-form must be a list")
Wrong type argument
```

That's it and the exit code is 255.

I have no idea where the eager macro-expansion failure is coming from and whether it's the cause of the error in the second case, and if so why it's not an error when using the outdated `ert`.

fbergroth:
@tarsius The eager macroexpansion failure seems to come from [this test](https://github.com/magnars/dash.el/blob/26c1bbe371db3cff89d4915488a55e381ee51fe5/dev/examples.el#L841).
I'm not sure how to rewrite it though, or if we can just remove it.

Fuco1:
I fear we might end up with "PHP-like" mess with virtually equivalent functions with different argument orders... it might get very frustrating for the users.

tarsius:
I am perfectly fine with this not being merged.

magnars:
That's okay. Thanks!

PR #113, opened 2015-01-13
Fix byte compilation test

fbergroth:
The old test fails because `emacs-snapshot` doesn't output `Wrote dash.elc`. This should be more robust.


magnars:
Thanks!

PR #114, opened 2015-01-13
Update error tests

fbergroth:
This fixes the error @tarsius ran into in #112.

On emacs24, eager macroexpansion triggered the signal in `-lambda` before actually running the test.
This also casused builtin ert to crash on emacs24+.

### DIFF
--- a/dev/examples-to-tests.el
+++ b/dev/examples-to-tests.el
@@ -3,9 +3,12 @@
 (require 'dash-functional)
 
 (defun example-to-should (example)
-  (let ((actual (car example))
-        (expected (nth 2 example)))
-    `(should (equal ,actual ,expected))))
+  (-let [(actual sym expected) example]
+    (cond
+     ((eq sym '=>)
+      `(should (equal ,actual ,expected)))
+     ((eq sym '!!>)
+      `(should-error (eval ',actual) :type ',expected)))))
 
 (defmacro defexamples (cmd &rest examples)
   `(ert-deftest ,cmd ()
--- a/dev/examples-to-tests.el
+++ b/dev/examples-to-tests.el
@@ -1,18 +1,22 @@
 (require 'ert)
-(require 'dash)
-(require 'dash-functional)
 
-(defun example-to-should (example)
-  (-let [(actual sym expected) example]
-    (cond
-     ((eq sym '=>)
-      `(should (equal ,actual ,expected)))
-     ((eq sym '!!>)
-      `(should-error (eval ',actual) :type ',expected)))))
+(defun example-to-should (actual sym expected)
+  (cond ((eq sym '=>)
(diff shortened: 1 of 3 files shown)


Fuco1:
Heh, good idea with the error operator.

Can you also change the other `condition-case' test to use it? (it's for `-let`). Bonus points for doing it in the same commit.

fbergroth:
Ah, missed that one. The PR has been updated.

dev/examples-to-tests.el
```diff
@@ -3,9 +3,12 @@
 (require 'dash-functional)
 
 (defun example-to-should (example)
-  (let ((actual (car example))
-        (expected (nth 2 example)))
-    `(should (equal ,actual ,expected))))
+  (-let [(actual sym expected) example]
```
magnars: Is it a good idea to have a test suite that is dependent on the code it is testing? :)

dev/examples-to-tests.el
```diff
@@ -3,9 +3,12 @@
...
+    (cond
+     ((eq sym '=>)
+      `(should (equal ,actual ,expected)))
+     ((eq sym '!!>)
```
magnars: This is a great addition! :+1:

dev/examples-to-tests.el
fbergroth: Not optimal, but probably ok I'd say. Besides, `-map` and `-partition` are already in there. :)

fbergroth: Or we could add tests for the test suite :smile:

tarsius:
So you guys really don't like `pcase`, or do you :-)
I personally think it's lovely.

fbergroth:
@tarsius that was my first approach, but it's not included on emacs23 :(

Fuco1:
@tarsius Personally, I find the `pcase` syntax quite horrible and over all it's an overkill for most applications.  In emacs core for example, it is never used for anything else than fancy "cond on strings" :/

tarsius:
I use it for to things.

* As a replacement for `cl-case`. The backticks don't really bother me here.
* When it really makes things easier. See e.g. `magit-apply.el`.

@Fuco1 I remember you saying that you disliked the syntax when `-let` was discussed. I can see what you mean, but I think it's okay. Which reminds me I have to look into `-let` I have completely missed that it has been merged, until I saw the changes to the tests above.

dev/examples-to-tests.el
Fuco1: @fbergroth Would you write this without `-let` so we can merge it? Better safe than sorry :)

fbergroth:
@Fuco1 @magnars now test suite doesn't use dash, and a couple of bad tests were found and fixed.

magnars:
Nicely done. Thanks! :-)

Issue #117, opened 2015-01-18
Shorthand lambda


fbergroth:
The anaphoric functions makes inline functions nice and concise, but `(lambda ..)` can feel a bit clunky to use elsewhere. What about a `-fn` (or `-f`) macro:
```el
(-fn "abc")     ;; => (lambda (&rest %&) "abc")
(-fn (* % %))   ;; => (lambda (%1 &rest %&) (* %1 %1)) -- % is translated to %1
(-fn (* %1 %2)) ;; => (lambda (%1 %2 &rest %&) (* %1 %2))
(-fn (* %2 %3)) ;; => (lambda (%1 %2 %3 &rest %&) (* %2 %3))
```
For extra sugar, `(-fn f x y)` could expand to `(-fn (f x y))` if `(functionp f)`.

Fuco1:
I would call it `-l` and have it expand into `-lambda` for extra extra sugar :) (if there's some sensible way to do it).

This would be excellent for use in nested `--` functions where the `it` can only refer to its own form, never outside.  Plus it solves the limitation to 2 arguments quite nicely (for example, for variadic `-zip)

fbergroth:
I can't think of a way to make `-l` use the destructuring of `-lambda`. I guess one can just throw in a `-let` somewhere :p. Or did you have something in mind?

fbergroth:
The `&rest` arg should be named `%&` like in clojure. (I forgot to mention I was trying to mimic clojure's anonymous function literal, in case someone is not familiar with that.)

Fuco1:
Are you working on this? There's a similar proposal to add this as core feature to elisp on the devel mailing list now.

I guess we can implement it until it gets in (which will probably be never) and then slowly phase out the anaphoric `--` variants which are quite ugly.

As I understand it, clojure uses `#(foo)` syntax, so what about `(-# foo)`?

fbergroth:
I'm not working on it, but it seems to be pretty trivial to implement. I also noticed it coming up on emacs-devel, but as you say, it will probably never get in.  :)

`-#` is a good one. My only annoyance is that `(-# fn arg1 arg2)` feels a bit hacky, but probably it's the best we can do. I might have some time to implement this during the weekend.

fbergroth:
Unfortunately `-#` isn't allowed by the reader. `-l` is a bit too close to `1-`, so maybe `-&`, `-%`, `-$`, or back to `-f`? :p

Fuco1:
See #118.

PR #118, opened 2015-01-23
Add short lambda macro

fbergroth:
Stab at what was discussed in #117. It looks a bit too weird and cryptic, I'm not sure I like it. :p


fbergroth:
Re-open if there's a prettier way to do this.

Issue #119, opened 2015-02-04
info manual


holomorph:
It would be nice if dash had an info manual.  Perhaps this could be achieved by:

* pandoc
* rewriting the skeleton in org, writing a new org generator and doing org -> info
* using some other format
* ???

Fuco1:
I don't oppose the change to org in principle, it depends (for me) on the fact whether github can render org as well as markup.

Autogenerating org or md markup is on par in difficulty (= very simple) and the skeleton can be rewritten by hand easily as well.

holomorph:
Played around and did the skeleton https://gist.github.com/holomorph/ed2ef265d3b9eb2dbf15

magnars:
Sounds like a fine idea to me. However, last time I tried it, github really did a horrible job of rendering org. That might have changed now, tho. Do a test if you can make it look good before dumping too much time into generating stuff.

Another option is to generate info-files in addition to the markdown. Create a new `examples-to-info.el` and go wild. An info-page wouldn't need the same exact content as is in the skeleton-file either, so making another one custom for info-pages sounds good too.

holomorph:
There does not appear to be any org analog to texinfo things like, for instance: `@table` for explaining commands (e.g. in the Installation section) and `@defun` for API function documentation, so without embedding texinfo into org (which will make github rendering fail miserably), org mode will not work.

I'm going to look into writing/generating texinfo instead.

holomorph:
Some generation and a little bit of poking later, here is an example of an info manual. You can load this into a Texinfo buffer and to `C-c RET C-b` to view it: https://gist.github.com/holomorph/3ffaa733a3ff02af4c96

Some things I noticed:
* in texinfo `@`, `{`, and `}` need to be escaped and there are two places where that's a problem: the `,@` operator and an example which has braces in it. Haven't quite figured out where to do the escaping
* some of the example docstrings have markdown quotes in them where they probably should be quoted differently
* Having a hard time making a node list like
```
* Maps::
* Sublist selection::
* List to list::
* Reductions::
* Unfolding::
* Predicates::
* Partitioning::
* Indexing::
* Set operations::
* Other list operations::
* Tree operations::
* Threading macros::
* Binding::
* Destructive operations::
* Function combinators::
```
from the function list

Fuco1:
Fixed via #120.

PR #120, opened 2015-02-07
Add info manual

holomorph:
Here's an info manual for #119

* The changelog is all in its own section instead of being divided into subsections
* The "Anaphoric functions" section is moved into "Functions" as an introduction. This is a change I'd make to the README because it makes more sense to introduce and explain the existence of the anaphoric macros before one runs into them in the function docs
* The function links are weird because there aren't any texinfo references that don't prepend "see" to the link. Only `@ref` does this _in the printed version_. Annoying.
* I probably did bad things™ in `examples-to-info.el`
* This is the same for README.md, but some of the really big examples--like the ones in the Trees section--end up making really longs lines, which is unfortunate

Anyways, I think the result is pretty nice. Let me know what you think.

### DIFF
--- a/create-docs.sh
+++ b/create-docs.sh
@@ -5,3 +5,4 @@ if [ -z "$EMACS" ] ; then
 fi
 
 $EMACS -batch -l dash.el -l dash-functional.el -l dev/examples-to-docs.el -l dev/examples.el -f create-docs-file
+$EMACS -batch -l dash.el -l dash-functional.el -l dev/examples-to-info.el -l dev/examples.el -f create-info-file
--- a/create-docs.sh
+++ b/create-docs.sh
@@ -4,5 +4,10 @@ if [ -z "$EMACS" ] ; then
     EMACS="emacs"
 fi
 
+if [ -z "$MAKEINFO" ] ; then
+    MAKEINFO="makeinfo"
+fi
+
 $EMACS -batch -l dash.el -l dash-functional.el -l dev/examples-to-docs.el -l dev/examples.el -f create-docs-file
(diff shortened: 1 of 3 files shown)


Fuco1:
Can we use `pp` to pretty print the long examples?

Fuco1:
The info template is now maintained as a separate file to the markdown template right? So on update, both need to be updated?

This can be a bit annoying, but all that changes there is mostly just changelogs and version information so it's not *that* bad.

holomorph:
Right, the info template is a separate file.

magnars:
Thanks for your work on this!

I get:

```
./create-docs.sh 
dash.texi:6: warning: unrecognized encoding name `UTF-8'.
```

Is that something to worry about?

Otherwise this looks pretty sweet. :)

holomorph:
It shouldn't be a problem. I imagine if the manual looks fine it's nothing to worry about.
Older texinfo didn't support UTF-8, but it's present and documented in texinfo 5.2 (I don't get the warning with 5.2)
```
The '@documentencoding' command declares the input document encoding.
Write it on a line by itself, with a valid encoding specification
following, near the beginning of the file but after '@setfilename'
(*note @setfilename::):

     @documentencoding ENC

  At present, Texinfo supports only these encodings:

'US-ASCII'
     This has no particular effect, but it's included for completeness.

'UTF-8'
     The vast global character encoding, expressed in 8-bit bytes.

...
```
A related ml post
https://lists.gnu.org/archive/html/bug-texinfo/2007-01/msg00014.html

magnars:
Goodie. Thanks :)

Fuco1:
I made a patch and regenerated the docs, but I'm not sure what info files to commit.  My texinfo is 5.1 and the thing in repo was generated on 4.8.  I got *tons* of changes.

Issue #123, opened 2015-03-10
`-fixfn` test hangs on OSX; reveals larger issue of float comparison


occidens:
Here is a interesting problem: I was running the test suite on my Mac, and it hung on the test of `-fixfn`. I then ran the tests in a Linux virtual machine with no problem.

Closer inspection reveals that there are some subtle platform-specific (or Emacs minor-version-specific?) differences in the floating-point math in the `cos` function. While the fixpoint of `cos` can be *verified* on both systems

```emacs-lisp
(= (cos 0.7390851332151607) 0.7390851332151607) ;; => t
```

On a Mac (tested on `GNU Emacs 24.4.1 x86_64-apple-darwin13.4.0, NS apple-appkit-1265.21`), `-fixfn` never converges on a solution because `cos` oscillates between *0.739&#x2026;8* and *0.739&#x2026;5*.

```emacs-lisp
(cos (cos 0.7390851332151608)) ;; => 0.7390851332151608

(cos (cos (cos 0.7390851332151608))) ;; => 0.7390851332151605

(cos (cos (cos (cos 0.7390851332151608)))) ;; => 0.7390851332151608
```

On Linux (tested on `GNU Emacs 24.4.91.1 (x86_64-unknown-linux-gnu, GTK+ Version 2.24.10)`), however, `cos` does converge to the value specified in the test suite.

```emacs-lisp
(cos (cos 0.7390851332151608)) ;; => 0.7390851332151607

(cos (cos (cos 0.7390851332151608))) ;; => 0.7390851332151607

(cos (cos (cos (cos 0.7390851332151608)))) ;; => 0.7390851332151607
```

As a temporary solution, I will submit a pull request in a moment for a change to `run-tests.sh` that will allow the test to be skipped.

To solve the larger issue, what do you think of the following solution, which would allow the caller of `-fixfn` to account for the specific nature of the function being analyzed?

-   Modify `-fixfn` to take two addtional optional arguments
    -   an equality test function, which defaults to `equal`, but which could
        be a fuzzy comparison of floats.
    -   an iteration limit function, which defaults to a simple counter
        with a default max value. The caller could alter the max value or
        specify some more sophisticated convergence test.
-   Modify the tests for `-fixfn` so that they use a fuzzy comparsion
    for floating point numbers such as the one given in the Emacs manual
    (with a narrower tolerance)
    
    ```emacs-lisp
    (defvar fuzz-factor 1.0e-15)
    (defun approx-equal (x y)
[...]

PR #124, opened 2015-03-11
Allow run-tests.sh to skip tests

occidens:
This a temporary fix on the way to a proper solution to #123, but I thought it would be of general use going forward, in case there are future tests that raise problems.

This change allows tests to be skipped for development purposes by setting the environment variable `ERT_SELECTOR' to a valid [ERT test selector](http://www.gnu.org/software/emacs/manual/html_node/ert/Test-Selectors.html#Test-Selectors), e.g.:
```
ERT_SELECTOR='(not "-fixfn")' ./run-tests.sh
```

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -6,13 +6,23 @@ if [ -z "$EMACS" ] ; then
     EMACS="emacs"
 fi
 
+# Run all tests by default.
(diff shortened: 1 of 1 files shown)


magnars:
Thank you!

PR #125, opened 2015-03-11
Add `-some`

### LINKED ISSUE #122: Equivalent of `cl-some`
`-any?` is a predicate returning a boolean. `-first` returns me the element, not the value of the "predicate" applied to it.  Is there anything better than `(car (-keep 'pred list))`?  If not, how would I go about implementing it?

wasamasa:
This is a port of the CL function `some` and the Scheme function `any`
from SRFI-1.  It is thought as addition to `-any?` (which returns a
boolean) and `-first` (which returns an element).  Unlike these it
returns the first truthy value of applying the predicate on each list
item.

Resolves #122.


magnars:
Excellent!

PR #129, opened 2015-03-14
docs

holomorph:
Applied the change I did to the info manual in #120 to the README (moving the Anaphoric functions section into the Functions section), and did a merge from the README since it was there.

I didn't update `dash.info` because it was done with texinfo 4.8, and I have texinfo 5.2 which does things differently.  For instance, either the default fill is different (or it just makes different decisions about filling paragraphs), uses fancy quotation marks, and uses unicode `⇒' instead of `=>', and of course basically any change to the .texi results in the references in the .info all being different.

I can understand keeping README.md in the repo, since the github repo is the homepage for the project, but keeping generated things in the repo (the generated .texi and .info) should be the exception, not the rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,31 @@ Include this in your emacs settings to get syntax highlighting:
 
 ## Functions
 
+All functions and constructs in the library are prefixed with a dash (-).
+
+There are also anaphoric versions of functions where that makes sense,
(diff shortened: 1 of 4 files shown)


Fuco1:
Thanks :)

PR #137, opened 2015-05-13
Doc update for --split-list.

phillord:
Just the result of running create-docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -397,6 +397,7 @@ See also: [`-splice`](#-splice-pred-fun-list), [`-insert-at`](#-insert-at-n-x-li
 ```el
 (-splice-list 'keywordp '(a b c) '(1 :foo 2)) ;; => '(1 a b c 2)
 (-splice-list 'keywordp nil '(1 :foo 2)) ;; => '(1 2)
+(--splice-list (keywordp it) '(a b c) '(1 :foo 2)) ;; => '(1 a b c 2)
 ```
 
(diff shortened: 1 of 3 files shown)


Fuco1:
It is currently unclear what to do with changes in the info file. Your patch produces over 2000 changes which are mostly cosmetic.  Since the output changes based on the texinfo version, we need to agree on one and use that to update the output.

Ideally, the script should not re-run texinfo on incompatible versions.

phillord:
All good questions. I added it because I ran create-docs.sh when making an independent change (as suggested by the docs) and I got back all of these changes (none stemming from my code modifications).

Personally, I would say remove the .info file entirely. I mean, in the idle world, remove the .texinfo file and the README.md as well, but I know why they are there. MELPA will create the .info file for you though, I think? So, why have it in the repo? Or am I missing something?

Fuco1:
Well, there are other repositories than MELPA.

The README.md is there to be displayed by github at the project/repo site.

The .texi file must be there, it's the source file from which you generate info.

phillord:
Yes, I agree -- as I said, I understand why the .texinfo and README are there.  But I don't understand why the .info file is there, and it is this is that is causing the large diff. If it is removed, MELPA will regenerate it. And, with marmalade, it can be generated by the uploader or their tool chain.

Still the immediate issue is that the doc in the repo is out of date wrt to the source and who ever runs create-docs.sh next is going to get a big diff.

Issue #142, opened 2015-05-30
Nested `when-let`s break edebug


cpitclaudel:
Pressing <kbd>C-u C-M-x</kdb> on the following definition causes the error shown below:

```elisp
(defun nested-when-lets ()
  (-when-let ((a (-when-let ((b 1)) 2))) 3))
```

The error does not occur if using the standard `when-let` in Emacs 25.0.

```
Debugger entered--Lisp error: (invalid-read-syntax "Expected form")
  signal(invalid-read-syntax ("Expected form"))
  edebug-syntax-error("Expected form")
  apply(edebug-syntax-error "Expected form")
  edebug-no-match((nil . 11016) "Expected form")
  apply(edebug-no-match (nil . 11016) "Expected form")
  edebug-top-element-required((nil . 11016) "Expected form")
  edebug-form((nil . 11016))
  edebug-match-form((nil . 11016))
  edebug-match-symbol((nil . 11016) form)
  edebug-match-one-spec((nil . 11016) form)
  edebug-match-specs((nil . 11016) (form) edebug-match-specs)
  edebug-match-specs((nil . 11016) (sexp form) edebug-match-specs)
  edebug-match-sublist((nil . 11016) (sexp form))
  edebug-match-list(((((b 1)) 2) (11009 (11010 (11011 . 11012) (11013 . 11014) . 11015) . 11016) (11017 . 11018) . 11019) (sexp form))
  edebug-match-one-spec(((((b 1)) 2) (11009 (11010 (11011 . 11012) (11013 . 11014) . 11015) . 11016) (11017 . 11018) . 11019) (sexp form))
  edebug-match-specs(((((b 1)) 2) (11009 (11010 (11011 . 11012) (11013 . 11014) . 11015) . 11016) (11017 . 11018) . 11019) ((sexp form) body) edebug-match-specs)
  edebug-match-sublist(((((b 1)) 2) (11009 (11010 (11011 . 11012) (11013 . 11014) . 11015) . 11016) (11017 . 11018) . 11019) ((sexp form) body))
  edebug-list-form(((((b 1)) 2) (11009 (11010 (11011 . 11012) (11013 . 11014) . 11015) . 11016) (11017 . 11018) . 11019))
  edebug-form((((-when-let ((b 1)) 2)) (10998 (10999 . 11008) (11009 (11010 (11011 . 11012) (11013 . 11014) . 11015) . 11016) (11017 . 11018) . 11019) . 11020))
[...]

PR #143, opened 2015-05-30
Make it explicit that not all VALS are evalled in -if-let*

cpitclaudel:
I was not initially sure whether something like

    (-when-let* ((a (subr-that-could-return-nil))
                 (b (blow-up-if-arg-is-nil a)))
      (message "All safe"))

would work properly. It does, so edit the docs (and that of `if-let*') to make it clear. Feel free to word it better of course (or to correct a misunderstanding of mine)

Initially mentioned in #142.


Fuco1:
I merged this manually after changing `evalled` to `evaluated`.  Thanks.

cpitclaudel:
Great! Thanks for dash.el :)

PR #144, opened 2015-06-11
Make ->>'s second arg optional

cammsaul:
Make the second arg to `->>` optional like `->`'s. This behavior is matches Clojure's `->>` and  `thread-last` in Emacs 25.

##### Before:
```emacs-lisp
(->> some-arg) ; error -- wrong number of args
```

##### After:
```emacs-lisp
(->> some-arg) ; -> some-arg
```

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1168,16 +1168,17 @@ second item in second form, etc."
                   (list form x)))
    (:else `(-> (-> ,x ,form) ,@more))))
 
-(defmacro ->> (x form &rest more)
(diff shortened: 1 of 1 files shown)


cammsaul:
I also simplified
```emacs-lisp
`(,(car form) ,@(cdr form) ,x)
```

to
```emacs-lisp
`(,@form ,x)
``` 

in the body of `->>`.

Fuco1:
Thanks, this is very reasonable :)

PR #145, opened 2015-06-12
Update ->> dox since second arg is now optional

camsaul:


Fuco1:
The docs are auto-generated from the code so this is now fixed (and I suppose has been for some time).  Either way, thanks for your contribution!

PR #146, opened 2015-06-14
Indicate macros in generated info manual

holomorph:


holomorph:
just fyi, this is so macros are actually listed as `Macro: ->` instead of `Function: ->`

Fuco1:
Heh... I should go over the PRs more often, slipped through my filters again... thanks :)

PR #147, opened 2015-06-15
Add edebug information to threading macros

davidshepherd7:
The various threading macros didn't contain a `declare` form specifying how the arguments are evaluated. This meant that they couldn't be used with the [Names](https://github.com/Malabarba/names/blob/master/TheNittyGritty.org#macros) package.

I've added declare forms which seem to work for [my case](https://github.com/davidshepherd7/electric-spacing/blob/per-mode-rule-lists/electric-spacing.el#L228) (all my tests pass when using Names), but I don't know edebug very well so it should probably be checked.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1161,6 +1161,7 @@ as `(nth i list)` for all i from INDICES."
 in the first form, making a list of it if it is not a list
 already. If there are more forms, insert the first form as the
 second item in second form, etc."
+  (declare (debug (form form &rest form)))
   (cond
    ((null form) x)
    ((null more) (if (listp form)
@@ -1173,6 +1174,7 @@ second item in second form, etc."
 in the first form, making a list of it if it is not a list
 already. If there are more forms, insert the first form as the
 last item in second form, etc."
+  (declare (debug (form form &rest form)))
   (cond
    ((null form) x)
    ((null more) (if (listp form)
@@ -1185,6 +1187,7 @@ last item in second form, etc."
 signified by the token `it' in the first form. If there are more
 forms, insert the first form at the position signified by `it' in
 in second form, etc."
+  (declare (debug (form form &rest form)))
   (if (null more)
       (if (listp form)
           (--map-when (eq it 'it) x form)


Fuco1:
I'm pretty sure it is impossible to have edebug work with threading macros, I've tried not once to implement it.

I don't know what Names does with it but did you try to actually run edebug on this?

davidshepherd7:
Ah ok, why is impossible?

I think Names just uses it to decide whether to treat the macro arguments as code or data. If the arguments are code then expands them with the namespace prefix.

I haven't tried using it with edebug so I'll give it a go.

davidshepherd7:
Starting with

    (defun foo (&rest args)
      (apply #'-sum args)
      (+ 1 2))

    (--> '(1 2 3)
         (identity it) 
         (foo it))

and instrumenting foo (with C-u 1 M-x eval-defun as suggested by the documentation) then eval'ing the sexp gives me a debugger inside foo. Then pressing T gives me the output (in `*Messages*`):

    Edebug: foo
    Trace fast...
    Result: -sum
    Result: ((1 2 3))
    Result: 6 (#o6, #x6, ?\C-f)
    Result: 3 (#o3, #x3, ?\C-c)
    3 (#o3, #x3, ?\C-c)


Instrumenting `-->` and running

    (--> 2
     (+ it 1) 
     (- it 10))

Then hitting T till it finishes gives me:

    Trace fast...
    Result: ((- it 10))
    Result: nil
    Result: 2 (#o2, #x2, ?\C-b)
    Result: (+ it 1)
    Result: ((- it 10))
    Result: (--> (--> 2 (+ it 1)) (- it 10)) [2 times]
    Trace fast...
    Result: nil
    Result: t
    Result: (- it 10)
    Result: t
    Result: (- it 10)
    Result: -
    Result: nil
    Result: it
    Result: t
    Result: (--> 2 (+ it 1))
    Result: 10 (#o12, #xa, ?\C-j)
    Result: nil
    Result: (- (--> 2 (+ it 1)) 10) [3 times]
    Trace fast...
    Result: nil
    Result: t
    Result: (+ it 1)
    Result: t
    Result: (+ it 1)
    Result: +
    Result: nil
    Result: it
    Result: t
    Result: 2 (#o2, #x2, ?\C-b)
    Result: 1 (#o1, #x1, ?\C-a)
    Result: nil
    Result: (+ 2 1) [3 times]
    -7 (#o377777777777777777771, #x3ffffffffffffff9)

which I don't completely understand but looks reasonable. You can at least see `-->` recursing to create a chain of `-->` calls and then the results being evaluated.

Do you want me to try out anything else?

Malabarba:
Back when I wrote Names I remember I tried to write debug specs for these macros but I figured out you can't have a working debug-spec for them while also helping Names (see below).

Firstly, let me say your example doesn't work for me. I get `Symbol's value as variable is void: it`. Maybe it's a difference in our Edebug versions (I'm on Emacs 25, FWIW).  
Secondly, trying to Edebug the other two macros (`->` and `->>`) with this also doesn't work (with `Wrong number of arguments: (3 . 3), 4`).

I think the proper debug specs for the first two macros (which is also what's used on Emacs 25's `thread-first/last`) is:

    (debug (form &rest [&or symbolp (symbolp &rest form)]))

The reason why I say this doesn't play well with Names is that there's no way to tell Names that the `symbolp` clauses correspond to function names (though I've thinking of just letting Names assume that by default).
Still, `dash` has nothing to do with that, and this is the correct way of writing these specs (I think).

I think the proper way of debugging `-->` is with the following spec

    (declare (debug (form &rest [&or symbolp (symbolp &rest [&or "it" form])])))

It worked on this example for me:

    (--> "test"
         (concat it " 2" (format "%s" current-prefix-arg))
         (concat it " 3"))

davidshepherd7:
> Firstly, let me say your example doesn't work for me.

Ok, I probably did something wrong then (first time trying out edebug).

I don't actually need this any more (I only had one call to a threading macro after I was done refactoring, so I decided to just remove it), so seeing as it doesn't work I'll just close this pull request. Thanks to both of you for looking into it though.

Fuco1:
@Malabarba Hm, that looks promising.  I tried a bit different strategy back in the day and I figured, because the macro changes the forms it can't really give you the result of the modified calls (it appends/prepends etc to the forms).  Now that's what macros normally do, but this one is also recursive and a bit messy and... I don't use them so I didn't care much either :D

I'm not terribly sure what edebug does so if this works I would be happy to get it in dash.

Malabarba:
>  I tried a bit different strategy back in the day and I figured, because the macro changes the forms it can't really give you the result of the modified calls 

Indeed, this macro is a little different in that you can't tell edebug to set a breakpoint *around* the forms, but you can still tell it to set brakpoints inside them.

> I'm not terribly sure what edebug does so if this works I would be happy to get it in dash.

It worked for me. :) And, the fact that it's used on the 25 threading macros makes me feel safer that it's correct. And besides, since it's just a debug-spec we know it's not gonna break any code.

PR #148, opened 2015-06-15
Add `-some->`, -some->>`, and `-some-->` macros.

cammsaul:
These are dash versions of the "nil-shortcutting" threading macros from Clojure, [some->](http://clojure.github.io/clojure/clojure.core-api.html#clojure.core/some-%3E) and [some->>](http://clojure.github.io/clojure/clojure.core-api.html#clojure.core/some-%3E%3E). Like `->`/`->>`/`-->`, but they only thread expressions through each successive form when they are non-nil.

```el
(->> 100
     (+ 20))
;; => 120

(->> nil
     (+ 20))
;; => Lisp error: (wrong-type-argument number-or-marker-p nil)

(-some->> nil
          (+ 20))
;; => nil
```

### DIFF
--- a/dev/examples.el
+++ b/dev/examples.el
@@ -733,7 +733,26 @@ new list."
   (defexamples -->
     (--> "def" (concat "abc" it "ghi")) => "abcdefghi"
     (--> "def" (concat "abc" it "ghi") (upcase it)) => "ABCDEFGHI"
-    (--> "def" (concat "abc" it "ghi") upcase) => "ABCDEFGHI"))
+    (--> "def" (concat "abc" it "ghi") upcase) => "ABCDEFGHI")
+
+  (defexamples -some->
+    (-some-> '(2 3 5)) => '(2 3 5)
+    (-some-> 5 square) => 25
+    (-some-> nil square) => nil
+    (-some-> 5 even? square) => nil)
+
+  (defexamples -some->>
+    (-some->> '(1 2 3) (-map 'square)) => '(1 4 9)
+    (-some->> '(1 3 5) (-last 'even?) (+ 100)) => nil
+    (-some->> '(2 4 6) (-last 'even?) (+ 100)) => 106
+    (-some->> '("A" "B" :c) (-filter 'stringp) (-reduce 'concat)) => "AB"
+    (-some->> '(:a :b :c) (-filter 'stringp) (-reduce 'concat)) => nil)
+
+  (defexamples -some-->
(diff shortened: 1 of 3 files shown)


magnars:
Nice, this is looking good.

Dash is heading into ELPA, which requires Free Software Foundation copyright assignment papers to be signed. Have you done so for Emacs? If not, fire off an email to assign@gnu.org

cammsaul:
Awesome!
I'll get the papers signed right away.

Fuco1:
@magnars does that mean we're going to abandon our glorious `-` prefix?

Also, I'm sort of getting lost in all the functionality which is being added to emacs core... there's `seq`, someone reimplemented the `if-let` things, threading is also there... the bad thing is that all of those solutions are totally inferior to dash., the good that maybe we should just replace it all with dash and be happy (now that they seem open to non-stupid APIs).

magnars:
No, it's just inclusion in ELPA, the official package repository, so that other packages in ELPA may use dash. We're not being embraced by the warmth of Emacs core, I'm afraid.

cammsaul:
@magnars I finally got my Emacs assignment papers accepted by the FSF, so this should be g2g

Fuco1:
I'm slowly going through my list of remaining fixes in order to push another release, will look at your PR somewhare in the process, if magnars is busy.

I plan to finish by the end of next week, hopefully.

dev/examples.el
```diff
@@ -733,7 +733,26 @@ new list."
...
     (--> "def" (concat "abc" it "ghi")) => "abcdefghi"
     (--> "def" (concat "abc" it "ghi") (upcase it)) => "ABCDEFGHI"
-    (--> "def" (concat "abc" it "ghi") upcase) => "ABCDEFGHI"))
+    (--> "def" (concat "abc" it "ghi") upcase) => "ABCDEFGHI")
+
+  (defexamples -some->
+    (-some-> '(2 3 5)) => '(2 3 5)
+    (-some-> 5 square) => 25
+    (-some-> nil square) => nil
+    (-some-> 5 even? square) => nil)
```
Fuco1: Only the first three examples are passed to the readme. Could you move this one up, or add some more when we see how the nil is handled? Because with just one form it's rather boring :)

Fuco1:
I fixed the issue and merged this manually.

cammsaul:
Thanks :+1:

Fuco1:
Thank *you* :)

PR #159, opened 2015-08-27
Add '-each-r' and '-each-r-while'.

doublep:
These are similar to `-each` and `-each-while` except they iterate list starting from the end. I have noticed that these would be useful at least for implementing (in a faster way) certain other `*-r` function in dash itself. Plus it's not that unheard of to want this functionality elsewhere.

* During iteration (using anaphoric macro, of course) `it-index` is bound to the values you'd expect: N - 1 ... 1, 0.

* List is not modified in any way, which is particularly important for the macro. E.g. this would give the expected result:

        (let ((l '("foo" "bar" "baz")))
          (--each-r l (message "there is string '%s' at position %d in list %s"
                               it it-index l)))

Implementation note: building vector is considerably faster than building a reversed list (vector takes less memory, so there is less GC), plus length comes naturally.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -93,6 +93,44 @@ Return nil, used for side-effects only."
 
 (put '-each-while 'lisp-indent-function 2)
 
+(defmacro --each-r (list &rest body)
+  "Anaphoric form of `-each-r'."
+  (declare (debug (form body))
+           (indent 1))
+  (let ((v (make-symbol "vector")))
+    `(let* ((,v (vconcat ,list))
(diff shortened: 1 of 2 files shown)


Fuco1:
Clever thing with the vector, but could you also add the note to the source itself (as a comment, not to the docstring).  If we're going for speed these "hacks" ought to be commented.

I can already thing of a use of this in one of my packages, so I'm for inclusion :) Makes the reverse phase a bit nicer.

doublep:
Done.

PR #160, opened 2015-08-28
Add support for GNU ELPA.

phillord:
Update the copyright lines following copyright assignment policy.

### DIFF
--- a/dash-functional.el
+++ b/dash-functional.el
@@ -1,6 +1,6 @@
 ;;; dash-functional.el --- Collection of useful combinators for Emacs Lisp  -*- lexical-binding: t -*-
 
-;; Copyright (C) 2013-2014 Matus Goljer, Magnar Sveen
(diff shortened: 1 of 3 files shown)


magnars:
:+1:

Fuco1:
Can we put that line there when the package isn't in emacs? Does all emacs code I write automatically belong to fsf when I have signed the asignment?

This seems like a weird thing to do.

phillord:
Matus, once we go live on ELPA, then the code is in emacs. At the moment, technically, this is not the case because it's not on ELPA, I guess, in the sense of having been released, although it's in the repo there. It's a bit of a bootstrap issue. Any code that you submitted to ELPA (or in Emacs core) becomes copyright FSF.

I was waiting for a new version of dash.el to come out before the code goes life on ELPA (i.e. not just in the git repo), as this avoids me having to do anything clever with git.

PR #161, opened 2015-09-14
Add -alter-with

cheunghy:
`-alter-with` alters implementation of function 1 with function 2 and evaluate body, after evaluation restore back.


magnars:
Hi! This looks like a useful function (in clojure it's called with-redefs), but I don't think it belongs in dash. Thanks tho!

Fuco1:
There's a package `noflet` which does something similar, as well as just regular `cl-flet`.

cheunghy:
@Fuco1 Thanks!

PR #164, opened 2015-10-05
Fix check_copyright; Silence compiler-warnings

phillord:
This is a variety of fixes from Stefan!

* .gitignore: Add files generated by "make" in elpa.git.

* dash.el (--map-first, --map-last): Silence compile warning when
`rep' does not make use of `it'.
(---partition-all-in-steps-reversed): Remove unused var `len'.
(dash--table-carry): Fix docstring and use DeMorgan.
(-find-indices): Remove unused var `i'.
(-compare-fn): Move before first use.

* dev/examples-to-tests.el:
* dev/examples-to-info.el:
* dev/examples-to-docs.el:
* dev/examples.el: Add copyright boilerplate.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -347,7 +347,7 @@ See also: `-map-when', `-replace-first'"
 
 (defmacro --map-first (pred rep list)
   "Anaphoric form of `-map-first'."
-  `(-map-first (lambda (it) ,pred) (lambda (it) ,rep) ,list))
+  `(-map-first (lambda (it) ,pred) (lambda (it) (ignore it) ,rep) ,list))
 
 (defun -map-last (pred rep list)
   "Replace first item in LIST satisfying PRED with result of REP called on this item.
@@ -357,7 +357,7 @@ See also: `-map-when', `-replace-last'"
 
 (defmacro --map-last (pred rep list)
   "Anaphoric form of `-map-last'."
-  `(-map-last (lambda (it) ,pred) (lambda (it) ,rep) ,list))
+  `(-map-last (lambda (it) ,pred) (lambda (it) (ignore it) ,rep) ,list))
 
 (defun -replace (old new list)
   "Replace all OLD items in LIST with NEW.
@@ -828,8 +828,7 @@ This function can be thought of as a generalization of
   "Private: Used by -partition-all-in-steps and -partition-in-steps."
   (when (< step 1)
     (error "Step must be a positive number, or you're looking at some juicy infinite loops."))
-  (let ((result nil)
-        (len 0))
+  (let ((result nil))
     (while list
       (!cons (-take n list) result)
       (setq list (-drop step list)))
@@ -1054,11 +1053,9 @@ element of LIST paired with the unmodified element of LIST."
 (defun dash--table-carry (lists restore-lists &optional re)
   "Helper for `-table' and `-table-flat'.
 
-If a list overflows, carry to the right and reset the list.
-
-Return how many lists were re-seted."
-  (while (and (not (car lists))
-              (not (equal lists '(nil))))
+If a list overflows, carry to the right and reset the list."
+  (while (not (or (car lists)
+                  (equal lists '(nil))))
     (setcar lists (car restore-lists))
     (pop (cadr lists))
     (!cdr lists)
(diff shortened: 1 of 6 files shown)


dash.el
```diff
@@ -1054,11 +1053,9 @@ element of LIST paired with the unmodified element of LIST."
 (defun dash--table-carry (lists restore-lists &optional re)
   "Helper for `-table' and `-table-flat'.
 
-If a list overflows, carry to the right and reset the list.
-
-Return how many lists were re-seted."
-  (while (and (not (car lists))
-              (not (equal lists '(nil))))
+If a list overflows, carry to the right and reset the list."
```
Fuco1: This seems useless.  I much prefer to read conditions as "this and that" then "not this or that", especially if I'm expecting a positive answer.

dash.el
```diff
@@ -1081,6 +1078,7 @@ order.  The dimension of the result is (length lists).
 See also: `-table-flat'"
   (let ((restore-lists (copy-sequence lists))
         (last-list (last lists))
+        ;; FIXME: Why not just use (make-list (length lists) nil)?
```
Fuco1: No reason, can be fixed.

magnars:
Looks like a good set of changes to me. I don't care much either way on the and-not / not-or issue.

Fuco1:
Just seems like a change just to make change.  But yes.

Another thing though, do we want to follow the GNU changelog convention? I find it quite useless (vc can tell you what files changed)... you should rather describe the idea behind the change than where it occured... the changes should be small enough to be easily reviewed with diff or what not.

I can accept function names in the commit as it makes grepping marginally simpler (but git can grep fulltext anyway).  What I would rather avoid is making function changes and things like adding headers in one commit.

magnars:
> Another thing though, do we want to follow the GNU changelog convention? I find it quite useless (vc can tell you what files changed)... you should rather describe the idea behind the change than where it occured... 

Agreed. I don't see how this commit breaks that?

> the changes should be small enough to be easily reviewed with diff or what not. 

> What I would rather avoid is making function changes and things like adding headers in one commit.

Yes, agreed on both accounts. I don't see it as such a big issue that this pull request must be closed tho. But certainly, for future changes, it would be preferable with several commits addressing distinct concerns.

dash.el
phillord: I guess that Stefan made that change because it's one boolean less, is all.

phillord:
I think I largely agree with you. Two commits (so the first line doesn't have a semi-colon in it!) would have made more sense to me.

From a practical POV, this commit is coming from the ELPA repo, and Stefan rather than directly from me. While, of course, I can add new commits to the PR straight-forwardly, rewriting them (by splitting it) is going to make my life a bit of a nightmare, at least as far as I understand git functionality. 

The same is true for the ChangeLog comment convention.

I'll push a fix for the FIXME after I've finished writing this.

Fuco1:
So, is it now going to be like... you will sync changes from elpa to here? Or we will push here and sync to elpa? I'm quite confused :o 

I have some patches lined up.

phillord:
In general, I expect that the majority of changes to dash will happen as they
have so far -- through github because that's what the dash developers are used
to. In this case, though, the change was made in the ELPA git because that is
what Stefan is used to.

I've pulled the changes from ELPA to my github clone, and then PR them to
yours. AFAICT, on my next pull from github and push to ELPA, git should
realise that the commit is already there and everything should be fine.

It should mean that I can keep the two repo's in sync without too many
problems. I'm very open to suggestions about better ways to do this, since
I've not done it before.

Sorry, I should have put all of this at the top. Rushed things a bit.

Fuco1:
Nah, in fact, we should probably open another thread to discuss the "procedure", or you can write something up and attach it to the readme/contributing section.  As some people might prefer to push to elpa directly?  Dunno.  I have no moral problems so far other than simply wanting to know how to handle things so we cause the minimum amount of trouble and confusion :)

phillord:
I'd certainly be grateful for that. It's also possible that I should have done the whole thing with git subtree.

magnars:
I'm merging this, but feel free to continue the discussion. :)

phillord:
Well, that all seems to have worked. I've pull recent changes (including the merge) from github and pushed to ELPA. I think we have a workflow:-)

magnars:
Great! :-)

PR #165, opened 2015-10-06
Prefer make-list over --map.

phillord:
This is a small performance improvement using the `make-list'
primitive directly, rather than the more complex use of number-sequence
and --map.


magnars:
Thanks!

PR #166, opened 2015-10-06
Add '-sort-keyed'.

doublep:
issue#156

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1888,6 +1888,32 @@ if the first element should sort before the second."
   (declare (debug (form form)))
   `(-sort (lambda (it other) ,form) ,list))
 
+(defun -sort-keyed (key-computer comparator list)
+  "Sort LIST by given keys, comparing them using COMPARATOR.
+First KEY-COMPUTER is used to compute key value for each list
+item.  Then resulting keys are sorted (stably) according to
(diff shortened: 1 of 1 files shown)


dash.el
```diff
@@ -1888,6 +1888,32 @@ if the first element should sort before the second."
...
+item.  Then resulting keys are sorted (stably) according to
+COMPARATOR.  Finally, a copy of the original LIST is built and
+rearranged in the same way.
+
+LIST is not modified by side effects.  KEY-COMPARATOR is called
+once for each element of LIST, i.e. with one argument.
+COMPARATOR is called with two keys, and should return non-nil if
+the first one (and thus its associated LIST element) should sort
+before the second."
+  (if (cdr list)
```
Fuco1: Is this really necessary? Map and sort on empty list returns nil, so that's not a problem, and while calling it on a list with length one is pointless, the performance hit is so negigible it is not worth the extra complexity.

Fuco1:
I think we should rename this to `-sort-on` so that it is a composition of `-sort` and `-on`, that would make it more clear I think.

Fuco1:
Dash now uses lexical binding so it should be safe.  How are you on time?  Do you want to finish this? :)

Issue #172, opened 2016-02-08
Add function -find-value analogous to -find-index


Wilfred:
Take a predicate pred and a list and return the the first element in the list satisfying the predicate, or nil if there is no such element.

E.g.

```
(-find-value #'odd? '(2 4 1 6 3 3 5 8)) ;; => 1
```

I'm happy to write a pull request for this. Have I missed a function that's doing this already?

Wilfred:
Argh, it's `-first`. I'll send a docs patch, as I totally missed this.

Fuco1:
We could possibly add an alias?

Wilfred:
PR is merged, closing.

PR #173, opened 2016-02-09
Add 'see also' to `-find-index` and `-find-last-index`.

Wilfred:
As discussed in #172 I find it easily to overlook `-first` when looking
at these functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -442,6 +442,8 @@ Return a new list of the items in `list` for which `pred` returns a non-nil valu
 
 Alias: `-select`
 
+See also: [`-keep`](#-keep-fn-list)
+
 ```el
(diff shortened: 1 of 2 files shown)


Wilfred:
The pre-commit hook has updated the manual and readme too. Let me know if you want me to take those out of the patch.

Fuco1:
I'm a bit afraid of the 2192 changes in the info file, mostly because each version of the compiler creates slightly different output... so we will get tons of these :/

Wilfred:
OK, I've removed the .info and .texi changes.

Fuco1:
We will have to figure out some way to update it though...maybe test for specific versions? Dunno.

Fuco1:
I re-run the CI, turns out the error was just because it couldn't grab the snapshot version temporarily.

Thanks.

Issue #176, opened 2016-04-22
Any plans for inclusion of the clojure "doto" macro?


sp3ctum:
The macro, as you must surely know, works like so:

```clojure
(doto x & forms)
Evaluates x then calls all of the methods and functions with the
value of x supplied at the front of the given arguments.  The forms
are evaluated in order.  Returns x.
 (doto (new java.util.HashMap) (.put "a" 1) (.put "b" 2))
```

https://clojuredocs.org/clojure.core/doto
https://github.com/clojure/clojure/blob/clojure-1.7.0/src/clj/clojure/core.clj#L3700

If you think this might be a good idea, I can try to make an elisp port.

Fuco1:
This is it I think

```
-> is a Lisp macro in `dash.el'.

(-> X &optional FORM &rest MORE)

Thread the expr through the forms. Insert X as the second item
in the first form, making a list of it if it is not a list
already. If there are more forms, insert the first form as the
second item in second form, etc.
```

sp3ctum:
It's almost the same, but `->` uses the result of each previous expression in the following expression.
The difference to `doto` is that doto always uses the first value.
So in the OP, `(.put (new java.utils.HashMap) "a" 1)` returns nil (void), and yet it's possible to use the second expression. You can think of them as modifying the hashmap.

With doto it's possible to use "setters" or side effects with less code.

Fuco1:
Oh, I see :) Java hashmap works by side effect :D  I'm getting too used to Haskell.

Very well, I guess we could add such a thing. Hopefully someone will steal it later and put it in the core.

If you want to try a PR, don't forget tests.  The readme is then autogenerated from docstring and the test-cases.

Fuco1:
We'll continue this in the PR.

PR #177, opened 2016-04-23
Add -doto macro (like Clojure's doto)

sp3ctum:
Here is its documentation in clojuredocs:

```
  (doto x & forms)

  Evaluates x then calls all of the methods and functions with the
  value of x supplied at the front of the given arguments.  The forms
  are evaluated in order.  Returns x.
  (doto (new java.util.HashMap) (.put "a" 1) (.put "b" 2))
```
Documentation:
https://clojuredocs.org/clojure.core/doto

Source (clojure):
https://github.com/clojure/clojure/blob/clojure-1.7.0/src/clj/clojure/core.clj#L3700

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -71,6 +71,19 @@ special values."
          (setq it-index (1+ it-index))
          (!cdr ,l)))))
 
+(defmacro -doto (eval-initial-value &rest forms)
+  "Take a target form, then many forms to call with the target form in the
+second position. Returns the target form."
+  (declare (indent 1))
+  (let ((retval (gensym)))
+    `(let ((,retval ,eval-initial-value))
+       ,@(mapcar (lambda (form)
+                   (if (sequencep form)
+                       `(,(-first-item form) ,retval ,@(rest form))
+                     `(funcall form ,retval)))
+                 forms)
+       ,retval)))
+
 (defun -each (list fn)
   "Call FN with every item in LIST. Return nil, used for side-effects only."
   (--each list (funcall fn it)))
--- a/dash.el
+++ b/dash.el
@@ -79,7 +79,7 @@ argument. Returns the target form."
     `(let ((,retval ,eval-initial-value))
        ,@(mapcar (lambda (form)
                    (if (sequencep form)
-                       `(,(-first-item form) ,retval ,@(rest form))
+                       `(,(-first-item form) ,retval ,@(cdr form))
                      `(funcall form ,retval)))
                  forms)
        ,retval)))
--- a/dash.el
+++ b/dash.el
@@ -72,8 +72,10 @@ special values."
          (!cdr ,l)))))
 
 (defmacro -doto (eval-initial-value &rest forms)
-  "Take a target form, then many forms to call with the target form as the first
-argument. Returns the target form."
+  "Evals a form, then inserts that form as the 2nd argument to other forms.
+The EVAL-INITIAL-VALUE form is evaluated once. Its result is
+passed to FORMS, which are then evaluated sequentially. Returns
+the target form."
   (declare (indent 1))
   (let ((retval (make-symbol "value")))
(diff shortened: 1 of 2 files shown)


sp3ctum:
This is also discussed briefly at #176

sp3ctum:
To be honest, I found it quite difficult to make good test cases. In Clojure I could have used a Java / Javascript object that would have `void` methods, as they are very common. But in elisp I found no such thing. If you can come up with an object that can be modified with side effects, I'll be happy to add a test case for it.

For reference, here is how I use this macro in my project:
```lisp
  (it "jumps out of the current sexp and enters evil-lispy-mode with )"
    (-doto (with-test-buffer "(expression| one)"
             (evil-insert-state)
             (ot--keyboard-input
              (ot--type ")")))
           (expect :to-be-in-lispy-mode)
           (expect :to-have-buffer-contents "(expression one)|")))
```
https://github.com/sp3ctum/evil-lispy/blob/master/tests/test-everything.el#L62

Here `expect` is a [buttercup](https://github.com/jorgenschaefer/emacs-buttercup) function that returns nil or signals an error. With doto, I'm able to write expectations that look very pretty (at least to me). :smile:

Fuco1:
There are hashtables you can modify by side effect in elisp, or use something with setcar/setcdr.


-------
As a side note, I have almost identical macro in `smartparens`, where you can group configuration of many modes with `sp-with-modes`, so that 

```
(sp-pair 'emacs-lisp-mode "(" ")")
(sp-pair 'clojure-mode "(" ")")
```
becomes
```
(sp-with-modes '(emacs-lisp-mode clojure-mode)
  (sp-pair "(" ")"))
```
which could equivalently be written as
```
(-doto '(emacs-lisp-mode clojure-mode)
  (sp-pair "(" ")"))
```

sp3ctum:
I added another example with `setcar & setcdr`. Hash-tables would have been nice, but it seems modification functions for them expect the table to be modified in the last position, and `-doto` always puts the object being modified as the first argument (second item).

sp3ctum:
Ping.
I'm in no rush with this, but in case you have more ideas I can work on them on my free time. Just let me know.

Fuco1:
Hi.  I think all is fine, except I think we should add a note about the evaluation in the docstring to make it crystal clear.  That is, the initial form is evaluated only once, then passed to each form which are then evaluated sequentially.

Also try to make the first line of the docstring shorter (less than 80 characters, end with period) so it will display properly in the apropos searches.  The arguments should also be described with at least one sentence each to please `checkdoc` :)

sp3ctum:
I finally made time to do this :)
I found the docstring hard to put into words. My hope is people will find the examples more useful. The clojure version's docstring is hard to read for me as well.

dash.el
```diff
@@ -71,6 +71,21 @@ special values."
          (setq it-index (1+ it-index))
          (!cdr ,l)))))
 
+(defmacro -doto (eval-initial-value &rest forms)
+  "Evals a form, then inserts that form as the 2nd argument to other forms.
```
Fuco1: Use imperative (Eval, insert) instead of 3rd person form.  That is the emacs standard.  In the rest of the explanation, 3rd person is fine.

I think this description is OK.  Once you do this last small edit, I'm merging! :+1:

sp3ctum: Ah, good point. Fixed in e5b3e4ac5caa45aed0fdb4331140272676918c35.

Fuco1:
Thanks! :)

sp3ctum:
Great!
I noticed the readme is still missing a mention of -doto. Want me to open a PR? I can run the generation script and commit the changes to the readme file.

Fuco1:
Oh yes, that's how we do it... you need to regen the file manually with the commit (a bit unfortunate it's not automaic somehow) Please do so, I'll rebase it in. Thanks!

PR #178, opened 2016-04-23
Add `-each-indexed` and an anaphoric equivalent

### LINKED ISSUE #175: Add -each-indexed analogous to -map-indexed
Allow users to iterate through a list, with an index, but without building a list of the results.

Wilfred:
This adds `-each-indexed` and an anaphoric variant `--each-indexed`,
equivalent to `-map-indexed`.

Whilst `--each` already defines `it-index`, it's not documented. Having
a separate version is more explicit and more discoverable because it's
similarly named to `-map-indexed`.

Fixes #175.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -77,6 +77,16 @@ special values."
 
 (put '-each 'lisp-indent-function 1)
 
+(defalias '--each-indexed '--each)
+
+(defun -each-indexed (list fn)
+  "Call (FN index item) for each item in LIST.
+
+In the anaphoric form `--each-indexed', the index is exposed as `it-index`.
+
+See also: `-map-indexed'."
+  (--each list (funcall fn it-index it)))
+
 (defmacro --each-while (list pred &rest body)
   "Anaphoric form of `-each-while'."
   (declare (debug (form form body))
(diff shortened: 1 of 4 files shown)


Fuco1:
`-map-indexed` takes a function of two arguments, `index` and `item`, so `-each-indexed` should too then to keep consistency.

Wilfred:
@Fuco1 OK, I've written a separate function for `-map-indexed`. I agree this is more consistent :)

Does this look sensible now?

dash.el
```diff
@@ -77,6 +77,14 @@ special values."
 
 (put '-each 'lisp-indent-function 1)
 
+(defalias '--each-indexed '--each)
+
+(defun -each-indexed (list fn)
+  "Call (FN index item) for each item in LIST.
+
+In the anaphoric form `--each-indexed', the index is exposed as `it-index`."
```
Fuco1: Maybe we could also add a See also section here pointing to `-map-indexed` (and viceversa)?

Wilfred: Done.

Fuco1:
:+1:

PR #180, opened 2016-05-14
Document -doto macro in the readme

sp3ctum:


sp3ctum:
After running here were lots and lots `create-docs.sh`, there were *a lot* of changes in `dash.info` and `dash.texi`. I think you had better run and review it yourself to make sure nothing incorrect gets committed.

Fuco1:
Hi.

You should probably reset your master and rebase on the upstream commit (which I squashed), because now you are sending us the patches "again".  Then put the documentation patch on a branch, that avoids the master conflicts in the future.

About the .info file, the problem is that different versions of texinfo generate different output... our solution is not to commit it for now... we will need to decide on some "supported" version.  .texi should be mostly fine, I'll check that.

sp3ctum:
This is now done in https://github.com/magnars/dash.el/pull/181

PR #186, opened 2016-05-31
Docstrings grammar fixes

MadAnd:


Fuco1:
Hi.

`unary` is pronounced `/ˈjunəri/`, with `j` being a semi-vowel it admits `a` not `an`.  The choice of the article is purely phonetic, as elision doesn't really happen on orthography level (cf. french `l'homme` where the `h` is silent).

The plural is a bug :)

MadAnd:
Hi @Fuco1,
thanks for nice insight. Since I'm not a native speaker, I had no idea that articles account for phonetic properties.

I have rewritten the commit, it now fixes two occurences of `an unary`.

Fuco1:
Thanks, this is super! :)

PR #171, opened 2016-01-24
Copyedit: argument name consistency

RubyTuesdayDONO:
Renamed argument `list1` to `list` in description to match function signature.

PR #187, opened 2016-06-05
Copyedit: argument name consistency

RubyTuesdayDONO:
Renamed argument `list1` to `list` in description to match function signature.
Followup on pull #171 (which I incorrectly submitted against the derived documentation).


Fuco1:
Thanks!

Issue #108, opened 2014-12-20
Sync ->/->> indentation with Clojure


bbatsov:
You know I'm fond of the current way `->/->>` are indented, but I feel that in the interest of compatibility (and the principle of least surprise) we should probably use the same indentation settings in dash.el as in clojure-mode.

PR #188, opened 2016-07-02
Edebug support for threading macros

Wilfred:
Given an elisp expression of the form:

    (->> (+ 1 1)
         (+ 2)
         (* 3)
         (format "%s"))

This tells edebug that the first argument is an expression, so users can
step through it. This is an improvement over the current situation,
where users cannot step through any part of a threaded expression.

I've also changed the indentation, so threading macros are indented
similarly to other scoping macros:

    (->> (+ 1 1)
      (+ 2)
      (* 3)
      (format "%s"))

Note that this indentation and debugging behaviour matches thread-first
in Emacs 25:
http://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/emacs-lisp/subr-x.el?id=emacs-25.0.95#n49

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1260,6 +1260,7 @@ See also: `-select-columns', `-select-by-indices'"
 in the first form, making a list of it if it is not a list
 already. If there are more forms, insert the first form as the
 second item in second form, etc."
+  (declare (debug (form &rest [&or symbolp (sexp &rest form)])))
   (cond
    ((null form) x)
    ((null more) (if (listp form)
@@ -1272,6 +1273,7 @@ second item in second form, etc."
 in the first form, making a list of it if it is not a list
 already. If there are more forms, insert the first form as the
 last item in second form, etc."
+  (declare (debug ->))
   (cond
    ((null form) x)
    ((null more) (if (listp form)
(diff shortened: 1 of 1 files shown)


Fuco1:
We had the indentation you propose before but it was changed to match clojure which is the model for this feature.

Given that `thread-first` now exist we should maybe consider backporting it and aliasing.  What do you think?

Wilfred:
That seems like a clean approach. The two macros are entirely equivalent.

Fuco1:
I think you can set indentation for alias different from the source function.  I would keep it in the clojure style as the macros are identical there.

Fuco1:
@bbatsov @magnars In #108 we changed the indentation to match clojure, now in emacs 25 the threading macros use the "before #108" style... this PR changes the indent to "emacs core" style.  What should we do?  I don't have any opinion on the matter, maybe slightly prefering the old way.

How should we go about using `thread-first` under the hood.  Conditionally alias it on e25+?

magnars:
> I don't have any opinion on the matter, maybe slightly prefering the old way. 

Same.

> How should we go about using thread-first under the hood. Conditionally alias it on e25+?

An alias sounds good.

Wilfred:
Thanks for the feedback. I've amended this PR to just deal with edebug.

We could certainly set up aliases etc in another PR too.

Fuco1:
Thanks.  This is good to merge then.

PR #193, opened 2016-08-20
Correct typo of 'does' to 'do' in -map-when docs

GuiltyDolphin:
Just noticed this when reading through.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -357,7 +357,7 @@ See also: `-each-indexed'."
        (nreverse ,r))))
 
 (defun -map-when (pred rep list)
-  "Return a new list where the elements in LIST that does not match the PRED function
+  "Return a new list where the elements in LIST that do not match the PRED function
 are unchanged, and where the elements in LIST that do match the PRED function are mapped
 through the REP function.
 


magnars:
Thanks! :+1:

dash.el
```diff
@@ -357,7 +357,7 @@ See also: `-each-indexed'."
        (nreverse ,r))))
 
 (defun -map-when (pred rep list)
-  "Return a new list where the elements in LIST that does not match the PRED function
+  "Return a new list where the elements in LIST that do not match the PRED function
```
Fuco1: We should probably either drop `where` or `that`, or change the `where` to `with`.  This way it makes no sense to me.  There should also be a period at the end.

GuiltyDolphin: So perhaps:

```
Return a new list with the same elements as LIST, but with any elements satisfying PRED first mapped through REP.
```

?

Fuco1: Actually, I confused this function with `-remove` :D The original is fine, sorry for the confusion.

(Though according to the emacs doc conventions, first line should be <80 char complete sentence... I don't see any way to describe this in such a short sentence :O)

GuiltyDolphin: Yes, it can be a little restrictive - especially considering verbosity (for clarity) is often encouraged in naming and descriptions.

PR #195, opened 2016-10-14
Feature/coverall integration

Fuco1:
* Remove emacs 23 tests
* Run tests using the evm & cask combo (as is standard these days)
* Run tests for all 24.* versions as well as 25.1
* Add coveralls support


magnars:
Looking good!

Fuco1:
Can't get undercover to work, it alawys gives me

    UNDERCOVER: No coverage information. Make sure that your files are not compiled?

I don't see anything that should compile dash before tests.  Hmm.

Fuco1:
So the problem was with doing `-l dash.el` etc... it has to be loaded via `require` in the code.  Still I can only pick up `dash-functional` and not `dash` itself.  I'm getting out of ideas :/

Wilfred:
Might I suggest separating out the EVM changes from moving to coveralls? AFAICS the EVM + cask work is ready to merge.

Fuco1:
That's a good idea.  I'm finishing next smartparens release now so I'll probably spend the weekend on that.  Do you want to try and get this separated?

basil-conto:
Is this still something we want to have?

PR #199, opened 2016-11-02
Add annotations to pure functions.

### LINKED ISSUE #198: Annotate functions as pure and side-effect free
These annotations help the byte-compiler do clever things like compile-time function execution. Since dash.el functions are often pure and side-effect free, this could be a good performance win.

I'm happy to whip up a patch.

Wilfred:
Adding the declaration `(pure t)` enables the byte-compiler to evaluate calls at compile-time (if all arguments are literals). Adding the declaration `(side-effect-free t)` allows the byte-compiler to remove calls entirely if the result is not used.

For functions that take a predicate as an argument, we cannot know whether that predicate is pure, so we do not annotate those.

We also don't annotate functions that depend on `-compare-fn`, as that may change at runtime.

Fixes #198.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -310,6 +310,7 @@ See also: `-remove', `-map-last'"
   "Remove all occurences of ITEM from LIST.
 
 Comparison is done with `equal'."
+  (declare (pure t) (side-effect-free t))
   (--remove (equal it item) list))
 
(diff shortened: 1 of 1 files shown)


Fuco1:
Awesome

magnars:
👏

PR #201, opened 2016-11-05
Add function signatures for aliases of built-in functions

Wilfred:
Otherwise, Emacs cannot show helpful signatures, and just uses ARG1 ARG2
instead.

For example, prior to this change, `-drop` was described as:

    -drop is an alias for `nthcdr' in `dash.el'.

    (-drop ARG1 ARG2)

    Return the tail of LIST without the first N items.


Fuco1:
Thanks!  I had no idea this doesn't work by default.

Wilfred:
Arguably it's an Emacs bug IMO.

Issue #196, opened 2016-10-15
permutation and power set


holomorph:
Hi, not sure if y'all care for more set logic or some combinatorics, but I had a need for these recently and thought I'd drop them here.  Of course, these would be a bit different written using dash.

```
(defun powerset (s)
  "Return the power set of S."
  (if (null s) '(())
    (let ((prev (powerset (cdr s))))
      (append (mapcar (lambda (elt) (cons (car s) elt)) prev)
              prev))))

(defun permutations (s)
  "Return the permutations of S."
  (if (null s) '(())
    (apply #'append
           (mapcar (lambda (x)
                     (mapcar (lambda (perm) (cons x perm))
                             (permutations (remove x s))))
                   s))))
```

Fuco1:
I was thinking about this on Monday.  I'm going through my notes and this has been sitting there for some time.  If you cook up a PR I'll be fine with including it.

PR #203, opened 2016-11-11
[Feature #196] Add -powerset and -permutations

holomorph:
a couple thoughts
- maybe application on empty list should just return nil, instead of a list containing nil
- perhaps there is a shortcut to writing these using dash functions/macros instead of what's there. I scratched my head a bit and couldn't think of one.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1959,6 +1959,22 @@ The test for equality is done with `equal',
 or with `-compare-fn' if that's non-nil."
   (--filter (not (-contains? list2 it)) list))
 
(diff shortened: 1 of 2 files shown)


holomorph:
@Fuco1 psst

Fuco1:
I think it should return `(nil)` to follow the mathematical definition.

In the permutation function the `remove` is a potential source of slowdown (O(n) for each x) but I would not worry much about performance here.  If it becomes a problem we can optimize later.

PR #205, opened 2016-11-18
Fix setf on emacs23

Wilfred:
`gv-define-simple-setter` is not defined on Emacs 23, so we need to use
the macro `defsetf` instead. `defsetf` is not autoloaded, so we must
require 'cl before using it.

However, the whole form is macro-expanded on all Emacs versions, so
writing:

```emacs-lisp
(eval-when-compile
  (if (fboundp 'gv-define-simple-setter)
      (gv-define-simple-setter -first-item setcar)
    (require 'cl)
    (defsetf -first-item (x) (val) `(setcar ,x ,val))))
```

produces byte-compile errors on Emacs 24+. The byte compiler there is
assuming that `(defsetf ...)` is a function call because it hasn't
executed the `(require 'cl)`.

We solve this by wrapping in another level of `eval-when-compile`. I
think this is roughly equivalent to meta levels:
http://www.phyast.pitt.edu/~micheles/scheme/scheme22.html

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -578,14 +578,26 @@ Alias: `-any'"
 
 \(fn LIST)")
 
-(gv-define-simple-setter -first-item setcar)
+;; TODO: emacs23 support, when dropped remove the condition
+(eval-when-compile
+  (require 'cl)
+  (if (fboundp 'gv-define-simple-setter)
+      (gv-define-simple-setter -first-item setcar)
+    (require 'cl)
+    (with-no-warnings
+      (defsetf -first-item (x) (val) `(setcar ,x ,val)))))
 
 (defun -last-item (list)
   "Return the last item of LIST, or nil on an empty list."
   (declare (pure t) (side-effect-free t))
   (car (last list)))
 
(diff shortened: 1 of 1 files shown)


Wilfred:
Other approaches I considered:

```
(eval-when-compile
  (require 'cl)
  (if (fboundp 'gv-define-simple-setter)
      (gv-define-simple-setter -first-item setcar)
    (defsetf -first-item (x) (val) `(setcar ,x ,val))))
```

This compiles cleanly, but unconditionally requiring cl seems like a letdown in dash, when our whole motivation is to avoid cl.

Actually, that gives me an idea.

Wilfred:
Bah, that passes on Emacs 23 and trunk, but not on Emacs 24. I'll have another go in a day or two.

Fuco1:
Unless you are doing this purely for fun of exploration, I would not want you to waste too much time on it.  As I said we're dropping e23 support so this is not *strictly* required to get the `gv` features in.

Wilfred:
OK, here's a fix. It's not as elegant or satisfying as the nested `eval-when-compile`, but it works and compiles cleanly.

PR #206, opened 2016-11-20
Correct use of eval-after-load

Silex:
With a symbol, only the file providing the feature triggers the
eval-after-load. With a string, any file named the same way triggers
it. This is problematic for custom configurations named after
packages.


magnars:
Thanks again!

Silex:
:wink:

PR #212, opened 2017-04-19
added alias `-fn' for `-lambda'.

wygulmage:
When you can't use anaphoric macros, it makes life 3 letters more pleasant. You don't seem to mind having lots of aliases to support different conventions, so I figured this would be acceptable.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1844,7 +1844,9 @@ has the usual semantics of `lambda'.  Furthermore, these get
 translated into normal lambda, so there is no performance
 penalty.
 
-See `-let' for the description of destructuring mechanism."
+See `-let' for the description of destructuring mechanism.
+
+Alias: `-fn'"
   (declare (doc-string 2) (indent defun)
            (debug (&define sexp
                            [&optional stringp]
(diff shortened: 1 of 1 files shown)


dash.el
```diff
@@ -1864,6 +1866,8 @@ See `-let' for the description of destructuring mechanism."
       `(lambda ,(--map (cadr it) inputs)
          (-let* ,inputs ,@body))))))
 
+(defmacro '-fn '-lambda)
```
basil-conto: I think you mean
```elisp
(defalias '-fn '-lambda)
```

Fuco1:
Hi and thanks for taking time helping out!

Personally, I'm not a huge fan of aliases as it fractures the API (subsets of users use different subsets of aliases which leads to unfamiliar code for the other people).

Originally, they were added to bridge conventions from clojure, haskell and lisp(cl) to make "an API for everyone", I suppose, but if we were to start again I would probably be against it.

dash.el
wygulmage: Yeah, you're absolutely right. Sorry about that brain fart.

wygulmage:
I understand and agree completely; honestly I was confused by the huge number of aliases in dash, but that's what gave me the courage to make the suggestion. I'll keep it as an alias in my own code and not try to force it on others.

Fuco1:
@wygulmage That seems like the best idea :) Maybe we'll clean it up for 3.0 as we're planning breaking changes anyway.

PR #215, opened 2017-04-25
Add -iota (#207)

### LINKED ISSUE #207: Consider adding a -range or -iota function wrapping or aliasing number-sequence from subr.el
It's nice to have all list functions in one place for discoverability.

`range` is the name for this function in [python](https://docs.python.org/3/library/functions.html#func-range) and [clojure](https://clojuredocs.org/clojure.core/range), and IMO is the better name.

`iota` is the name in [APL, Go](https://en.wikipedia.org/wiki/Iota) [C++](http://en.cppreference.com/w/cpp/algorithm/iota) and [scheme](http://stackoverflow.com/questions/12909856/analog-of-pythons-range-in-scheme).

I can't find if this function exists in Haskell (I guess people just use the shorthand notation) or common lisp.

holomorph:
Fixes #207

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -2154,6 +2154,23 @@ N is the length of the returned list."
         (push (funcall fun (car r)) r))
       (nreverse r))))
 
+(defun -iota (count &optional start step)
+  "Return a list containing COUNT numbers.
+Starts from START and adds STEP each time.  The default START is
+zero, the default STEP is 1.
+This function takes its name from the corresponding primitive in
+the APL language."
+  (if (or (not (integerp count)) (< count 0))
+      (signal 'wrong-type-argument count))
+  (if (and step (zerop step)) (make-list count start)
+    (let ((res '())
+          (x (or start 0))
+          (dx (or step 1)))
+      (while (<= 0 (setq count (1- count)))
+        (push x res)
+        (setq x (+ x dx)))
+      (nreverse res))))
+
 (defun -fix (fn list)
   "Compute the (least) fixpoint of FN with initial input LIST.
 
--- a/dash.el
+++ b/dash.el
@@ -2160,16 +2160,16 @@ Starts from START and adds STEP each time.  The default START is
 zero, the default STEP is 1.
 This function takes its name from the corresponding primitive in
 the APL language."
-  (if (or (not (integerp count)) (< count 0))
-      (signal 'wrong-type-argument count))
-  (if (and step (zerop step)) (make-list count start)
-    (let ((res '())
-          (x (or start 0))
-          (dx (or step 1)))
+  (when (not (natnump count))
+    (signal 'wrong-type-argument (list #'natnump count)))
+  (or start (setq start 0))
+  (or step (setq step 1))
+  (if (zerop step) (make-list count start)
+    (let (result)
       (while (<= 0 (setq count (1- count)))
-        (push x res)
-        (setq x (+ x dx)))
-      (nreverse res))))
+        (push start result)
+        (setq start (+ start step)))
+      (nreverse result))))
 
 (defun -fix (fn list)
   "Compute the (least) fixpoint of FN with initial input LIST.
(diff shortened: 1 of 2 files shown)


Fuco1:
I'm torn between `-iota` and `-range`... scheme is fine but then we also use a lot of clojure conventions.

I'll leave this open for a bit, maybe someone will have some constructive comments.

But otherwise good work, this is good to merge once we decide the name.

magnars:
Since the dash API is heavily inspired by the Clojure core API, I think it should follow that convention. Having to make excuses for the function name in the docstring is a clue too. ;-)

Fuco1:
@magnars reading the clojure api docs (https://clojuredocs.org/clojure.core/range) it seems to have different semantics.  Clojure uses `start end step` like `(number-sequence FROM &optional TO INC)` in elisp, but this function here uses `count start step` as arguments.

range in python also uses `start stop step` convention.

I haven't realized that before :blush: So I think this shouldn't be called range to avoid confusion over the arguments.

Maybe we could add another function `-range` aliasing to `number-sequence`.

dash.el
```diff
@@ -2154,6 +2154,23 @@ N is the length of the returned list."
...
       (nreverse r))))
 
+(defun -iota (count &optional start step)
+  "Return a list containing COUNT numbers.
+Starts from START and adds STEP each time.  The default START is
+zero, the default STEP is 1.
+This function takes its name from the corresponding primitive in
+the APL language."
+  (if (or (not (integerp count)) (< count 0))
+      (signal 'wrong-type-argument count))
```
holomorph: Added one

Fuco1:
So I'm thinking we add `-range` as an alias to `number-sequence` and leave this one as it is.  Then we should add references to each other in the docstrings so that people can look around and see how they differ.

@magnars what do you think?

magnars:
That sounds OK to me.

holomorph:
Ping? :P

dash.el
```diff
@@ -2154,6 +2154,23 @@ N is the length of the returned list."
         (push (funcall fun (car r)) r))
       (nreverse r))))
 
+(defun -iota (count &optional start step)
+  "Return a list containing COUNT numbers.
+Starts from START and adds STEP each time.  The default START is
+zero, the default STEP is 1.
+This function takes its name from the corresponding primitive in
+the APL language."
+  (if (or (not (integerp count)) (< count 0))
```
basil-conto, line 2163: AKA `(not (natnump count))`.

dash.el
basil-conto, line 2164: `signal` takes a list as second argument:

```el
(signal 'wrong-type-argument (list #'natnump count))
```

dash.el
```diff
@@ -2154,6 +2154,23 @@ N is the length of the returned list."
...
+      (signal 'wrong-type-argument count))
+  (if (and step (zerop step)) (make-list count start)
```
basil-conto, line 2165: What if `start` is `nil`? According to the docstring, I don't think this is right:

```el
(-iota 2 nil 0) ; => (nil nil)
```

dash.el
holomorph: Pretty sure I wrote it this way because natnup lacks a bytecode

dash.el
holomorph: Thanks, good catch--will have to fix that.

dash.el
basil-conto: So? It's still faster to only call 2 functions rather than 3 plus a special form.

dash.el
basil-conto: I think it should be as simple as hoisting

```el
(or start (setq start 0))
(or step (setq step 1))
```
before the `zerop` check.

holomorph: Agreed, I think that's the simplest way to address it.

basil-conto (approved):

basil-conto:
Thanks!

PR #216, opened 2017-04-25
Replace string-to-int with string-to-number

holomorph:
string-to-int was declared obsolete in Emacs 22.1, and has been removed
for the eventual 26.1 release.


Fuco1:
Huh, `string-to-int` returns "Return 0 if STRING cannot be parsed as an integer or **floating** point number."  What the hell :D

Thanks for the patch, I guess there's not much to discuss.

PR #220, opened 2017-05-23
Fix -map-last docs

AdamNiederer:
This is a copy-paste typo, right?


Fuco1:
Yep, thanks for catching this :)

PR #221, opened 2017-05-25
Make --> bind IT for use anywhere in FORMS, and add -as->.

### LINKED ISSUE #218: --> doesn't insert `it` if `it` isn't at the top level of a form
I'm using version 2.13.0 of dash.

I had a call to `-->` where I inserted `it` as not the top-level element of a form, but a secondary one. Sample code:

    (--> 3
         (car (list it)))

I would expect this code to return `3`, but it instead errors:

    *** Eval error ***  Symbol’s value as variable is void: it

It works if I only insert `it` at the top-level of the forms:

    (--> 3
         (list it)
         (car it))

zck:
This code fixes #218, in a way that I think is fairly straightforward, and not error-prone. I've added a example to show the problem that was fixed. 

It keeps the feature that if a FORM is a symbol, the code will treat that symbol as a function, and pass it the single arg `it`. I keep this only because it's a legacy feature; it's [not how Clojure operates](https://repl.it/IGM4/1), and was surprising to me. But, as mentioned, I kept it.

This also fixes a multiple-evaluation bug; if `it` appears multiple times in a FORM, the value will be evaluated multiple times. The following code prompts for input twice with the old `-->`, but only once with this version.

`(--> (progn (read-from-minibuffer "How many prompts? ") 3) (+ it it))`

As a helper method, this adds a new method `-as->`, which is a version of `-->` that takes the variable explicitly, as opposed to `-->` only using `it`. I think the standard dash.el nomenclature would be to name the anaphoric version with an extra dash, but the anaphoric version is `-->`, and the version with one fewer dash (`->`) is already used. Please let me know if there's a better name for this new function. The existing name is a riff on Clojure's, but beginning with a dash to fit in with names here.

I'm happy to make changes, whether to the name, or anything else here.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1354,17 +1354,29 @@ last item in second form, etc."
                   (list form x)))
    (:else `(->> (->> ,x ,form) ,@more))))
 
-(defmacro --> (x form &rest more)
-  "Thread the expr through the forms. Insert X at the position
-signified by the token `it' in the first form. If there are more
-forms, insert the first form at the position signified by `it' in
-in second form, etc."
-  (declare (debug (form &rest [&or symbolp (sexp &rest [&or "it" form])])))
-  (if (null more)
-      (if (listp form)
-          (--map-when (eq it 'it) x form)
-        (list form x))
-    `(--> (--> ,x ,form) ,@more)))
+(defmacro --> (x &rest forms)
+  "Starting with the value of X, thread each expression through FORMS.
+
+Insert X at the position signified by the token `it' in the first
+form.  If there are more forms, insert the first form at the position
+signified by `it' in in second form, etc."
+  (declare (debug (form body)))
+  `(-as-> ,x it ,@forms))
+
+(defmacro -as-> (value variable &rest forms)
+  "Starting with VALUE, thread VARIABLE through FORMS.
+
+In the first form, bind VARIABLE to VALUE.  In the second form, bind
+VARIABLE to the result of the first form, and so forth."
+  (declare (debug (form symbolp body)))
+  (if (null forms)
+      `,value
+    `(let ((,variable ,value))
+       (-as-> ,(if (symbolp (car forms))
+                 (list (car forms) variable)
+               (car forms))
+            ,variable
+              ,@(cdr forms)))))
 
 (defmacro -some-> (x &optional form &rest more)
   "When expr is non-nil, thread it through the first form (via `->'),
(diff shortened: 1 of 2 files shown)


zck:
Also to mention: I think I have the `declare`statements right, but am not sure. I was able to debug into the macros, but I have very little experience using edebug, so I could very well have done it incorrectly and not noticed, because I'm not sure how edebug is supposed to work.

zck:
I'm not sure why, but the Travis CI build [has finished successfully](https://travis-ci.org/magnars/dash.el/builds/235897989?utm_source=github_status&utm_medium=notification), but hasn't reported back here.

Fuco1:
I marked this for a review.  I played a bit around and couldn't find a counter example where it would break but I'm still kind of afraid that some of the anaphora macros will expand in a stupid way and things will get ugly.  Going over each one is out of the question I guess, that would be just too much work.

Or maybe we can just try it and see... I hope nobody's powering spacecraft with Emacs :P

zck:
That's fair. I might go through and add some more examples using more complicated anaphoric macros. 

Also I tried to find a space game or something in Emacs to make a joke about, but couldn't find one. I'm pretty disappointed, to be honest.

zck:
Here are some examples mixing `-->` with other anaphoric operators. They're not good examples of using the dash library, so I don't think we should add them to the examples, but they're useful for seeing if the macros have problems with multiple `it`s. Although if you prefer to add them as tests, let me know and I'll do that. 

    ELISP> (--> '(1 2 3 4) (--map (1+ it) it))
    (2 3 4 5)
    ELISP> (--map (--> it (1+ it)) '(1 2 3 4))
    (2 3 4 5)

    ELISP> (--filter (--> it (equal 0 (mod it 2))) '(1 2 3 4))
    (2 4)
    ELISP> (--> '(1 2 3 4) (--filter (equal 0 (mod it 2)) it))
    (2 4)

    ELISP> (--annotate (--> it (< 1 it)) '(0 1 2 3))
    ((nil . 0)
     (nil . 1)
     (t . 2)
     (t . 3))
    
    ELISP> (--> '(0 1 2 3) (--annotate (< 1 it) it))
    ((nil . 0)
     (nil . 1)
     (t . 2)
     (t . 3))

Fuco1:
Yes please, add them as tests and then we merge.

Only the first three examples are picked for the documentation so if we don't want to show these add three "good ones" and place these after that.

zck:
I added the tests.

I added them as a separate commit, but I can squash them if you prefer.

Fuco1:
No problem!  Thanks for your contribution!

PR #222, opened 2017-06-01
Add new methods to partition a list each time elements occur, or a predicate is true, but include those elements in the list.

zck:
These methods partition a list each time a predicate is true, or an item is found. These are similar to `-split-on` and `-split-when`, but include the item that necessitates the split. These are also similar to `-partition-by` and other similar partition functions. Examples/tests are included.

The new methods are -partition-before-pred, -partition-after-pred, -partition-before-item, and -partition-after-item. I didn't name them with `-on` (for items) or `-when` (for predicates) because I couldn't find a way to include that in the name and also include `-before` and `-after` without sounding awkward. I'm happy to do a rename if you have ideas for better names.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1032,6 +1032,36 @@ returns the header value, but only after seeing at least one
 other value (the body)."
   (--partition-by-header (funcall fn it) list))
 
+(defun -partition-after-pred (pred list)
+  "Partition directly after each time PRED is true on an element of LIST."
+  (when list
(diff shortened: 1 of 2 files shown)


zck:
Is there anything I can do to make this easier to evaluate?

Fuco1:
It seems fine.  I think nobody noticed the PR :( Sorry for keeping this open for so long.

zck:
No worries. Thanks for merging it in!

PR #223, opened 2017-07-11
Add copyright assignment statement to documentation

phillord:
I'm about to start the process of updating dash in ELPA (rather long delayed); I'm going to need to get a few more copyright assignments done, but I wanted to add a statement about this to the documentation.

I ran create-doc -- it seemed to make some extra changes as well. Do you want me to separate out these changes into a separate PR?


Fuco1:
Does dash now require assignment? :O

phillord:
@Fuco1 I talked with @magnars about this a long time back, when I worked to get dash into ELPA for which copyright assignments are needed. I should have put a statement up then, but didn't. Now, I find myself in the position of asking a set of new contributors, so I can update the version on ELPA which is old.

Actually, I discussed this with you previously as well: https://github.com/magnars/dash.el/pull/160

Fuco1:
Ah well, I tend to forget :)  To me this seems like a bad idea but I don't have a strong opinion.  At this stage the project is stable and not much code gets contributed anymore.

phillord:
Which is a bad idea? Adding it to ELPA? Or the copyright assignment.

Fuco1:
Well they go hand in hand.  The CA is a problem for many people, for some because of their geopolitical situation, for others philosophically (some people don't really like to support FSF/Emacs in their current form... as "toxic" institutions whatever that means).  But because the project does not attract many contributors I don't see it as particularly bad *for the state of the project* right now.

I wouldn't do it with smartparens for example where I get quite a lot of patches from a lot of people for whom it would be an extra needless annoyance.

phillord:
I thought that this would be a problem also when I first undertook to get dash into ELPA. But, actually, it wasn't. No one objected, although some were not contactable.

I'm trying to work with the FSF to make the process smoother where ever we can. We see how it goes.

Fuco1:
@magnars I'm ok with the merge, shall we go ahead?

magnars:
Sounds good to me.
ons. 12. jul. 2017 kl. 12.44 skrev Matus Goljer <notifications@github.com>:

Fuco1:
@phillord thanks for the readme update as well :)

PR #232, opened 2017-08-16
Docs: Clarify that we don't modify the original list

### LINKED ISSUE #231: -remove-at does not match its docs
`-remove-at` is grouped under "Bag of various functions which modify input list."

However, it doesn't actually modify the list:

```emacs-lisp
(setq wh/x (list 1 2 3 4 5))
(-remove-at 1 wh/x)
wh/x ;; => (1 2 3 4 5)
```

Is this section of functions just general list-to-list functions, or is the function in the wrong section?

Wilfred:
Fixes #231

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ Functions returning a sublist of the original list.
 ### List to list
 
 
-Bag of various functions which modify input list.
+Functions returning a modified copy of the input list.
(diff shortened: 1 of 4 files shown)


Fuco1:
Cool, thanks!

PR #237, opened 2017-09-24
Add 'see also' for -select and -remove

Wilfred:
These functions do the opposite of each other, so it's useful to
mention the other one in the docstrings.


Wilfred:
Fixed merge conflicts.

PR #238, opened 2017-09-24
Define -second-item through to -fifth-item

Wilfred:
cl-second is a very useful function that has no dash
equivalent. Define functions for accessing the second, third, fourth
and fifth items of a list.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -582,12 +582,42 @@ Alias: `-any'"
 (defalias '-first-item 'car
   "Return the first item of LIST, or nil on an empty list.
 
+See also: `-second-item', `-last-item'.
+
 \(fn LIST)")
 
 ;; Ensure that calls to `-first-item' are compiled to a single opcode,
 ;; just like `car'.
 (put '-first-item 'byte-opcode 'byte-car)
 (put '-first-item 'byte-compile 'byte-compile-one-arg)
+
+(defalias '-second-item 'cadr
+  "Return the second item of LIST, or nil if LIST is too short.
+
+See also: `-third-item'.
+
+\(fn LIST)")
+
+(defalias '-third-item 'caddr
+  "Return the third item of LIST, or nil if LIST is too short.
+
+See also: `-fourth-item'.
+
+\(fn LIST)")
+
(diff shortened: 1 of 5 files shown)


dash.el
```diff
@@ -576,13 +576,43 @@ Alias: `-any'"
...
+\(fn LIST)")
+
+(defalias '-third-item 'caddr
+  "Return the third item of LIST, or nil if LIST is too short.
+
+See also: `-fourth-item'.
+
+\(fn LIST)")
+
+(defalias '-fourth-item 'cadddr
```
Fuco1: Cadddr comes from `cl` which we don't use.  Is it built-in?  I'm not sure.

I'm not sure if copying the definition would be better though, e.g.

```elisp
(defun cl-cadddr (x)
  "Return the `car' of the `cdr' of the `cdr' of the `cdr' of X."
  (declare (compiler-macro cl--compiler-macro-cXXr))
  (car (cdr (cdr (cdr x)))))
``` 

could be just copied into dash.

Wilfred:
Rebased and fixed as suggested. ✨

Fuco1:
It would also be cool if we made them into generalized variables so we could use `setf` with them.  But that's probably for a separate PR.  I'm going to merge this one, thanks!

Fuco1:
Oh this info thing is hell... I think we'll need to find a better workflow than people changing the file in PRs... ideally some bot or CI could do that and push an update.

Wilfred:
OK, rebased :)

I think dash.el would be more convenient if it didn't set up the commit hooks. Creating a PR without updating dash.info is fine (you could always generate it after), but getting a merge conflict is a little more hassle for new contributors.

Fuco1:
Does it setup the commit hooks though?  I don't have any.  I think you have to do it manually somehow.

Wilfred:
You're right. Looks like I followed the instructions in the readme months ago:

> I highly recommend that you install these as a pre-commit hook, so that
the tests are always running and the docs are always in sync:
>
>      cp pre-commit.sh .git/hooks/pre-commit

PR #240, opened 2017-09-25
Replace -first with -some in -any?

### LINKED ISSUE #239: Problem with any/some/all/every family of functions.
This gives the wrong answer: `(-any-p 'null '(nil))`

I believe the problem will manifest for any predicate that can return non-nil for nil.

basil-conto:
* Fixes #239 
* Extend examples accordingly
* Update docs


magnars:
Thanks!

PR #253, opened 2017-11-13
Add -keep-first

### LINKED ISSUE #252: -keep-first: Return the first result for an x in LIST where (PRED x) is non-nil, else nil.
I often with there was a form that is a combination of `-keep` and `-first`.

* `-keep` Return a new list of the non-nil results of applying fn to the items in list.
* `-first` Return the first x in list where (pred x) is non-nil, else nil.
* **`-keep-first`** Return the first result for an x in list where (pred x) is non-nil, else nil.

Does that already exist and I just cannot find it?

tarsius:
I had a quick stab at it.

Closes #252.

### DIFF
--- a/dev/examples.el
+++ b/dev/examples.el
@@ -746,6 +746,10 @@ new list."
     (-first 'null '(1 3 ())) => nil
     (--first (> it 2) '(1 2 3)) => 3)
 
+  (defexamples -keep-first
+    (-keep-first (lambda (n) (and (> n 1) (* n 1.0))) '(1 2 3)) => 2.0
+    (--keep-first (and (even? (car it)) (cdr it)) '((1 . "one") (2 . "two"))) => "two")
+
   (defexamples -some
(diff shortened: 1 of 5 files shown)


dev/examples.el
```diff
@@ -746,6 +746,10 @@ new list."
     (-first 'null '(1 3 ())) => nil
     (--first (> it 2) '(1 2 3)) => 3)
 
+  (defexamples -keep-first
+    (-keep-first (lambda (n) (and (> n 1) (* n 1.0))) '(1 2 3)) => 2.0
+    (--keep-first (and (even? (car it)) (cdr it)) '((1 . "one") (2 . "two"))) => "two")
```
basil-conto: Where is `even?` defined?

tarsius: At the top of this file.

basil-conto: As promised, thus delivered. :)

PR #257, opened 2017-12-31
Regenerate docs

Wilfred:
I've noticed that the info manual was outdated, so I've regenerated it.


Fuco1:
Ugh, we really should add a CI check to see if docs are up-to-date.  Thanks!

PR #258, opened 2017-12-31
Ensure ' in docstring code snippets is not replaced

Wilfred:
As of Emacs 25.1, `foo' is displayed as ‘foo’. This is controlled by
text-quoting-style, but curved quotes are the default.

This causes problems for docstrings that contain ' in code
snippets. Help buffers show e.g. (funcall 'foo) as (funcall ’foo),
which isn't legal elisp.

Instead, escape the quotes with `\=`. See substitute-command-keys for
the full docs on `\=` escaping.

### DIFF
--- a/dash-functional.el
+++ b/dash-functional.el
@@ -207,10 +207,10 @@ In types (for n=2): ((a -> b), (c -> d)) -> (a, c) -> (b, d)
 
 This function satisfies the following laws:
 
-  (-compose (-prodfn f g ...) (-prodfn f' g' ...)) = (-prodfn (-compose f f') (-compose g g') ...)
-  (-prodfn f g ...) = (-juxt (-compose f (-partial 'nth 0)) (-compose g (-partial 'nth 1)) ...)
-  (-compose (-prodfn f g ...) (-juxt f' g' ...)) = (-juxt (-compose f f') (-compose g g') ...)
-  (-compose (-partial 'nth n) (-prod f1 f2 ...)) = (-compose fn (-partial 'nth n))"
+  (-compose (-prodfn f g ...) (-prodfn f\\=' g\\=' ...)) = (-prodfn (-compose f f\\=') (-compose g g\\=') ...)
+  (-prodfn f g ...) = (-juxt (-compose f (-partial \\='nth 0)) (-compose g (-partial \\='nth 1)) ...)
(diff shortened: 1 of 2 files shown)


Fuco1:
Hm.  I don't like this at all.  It seems silly to turn one into another :(  If people want curly quotes why not put curly quotes in the string?  I understand the appeal to have it consistent, but this can mess with quite a lot of documentation.  At least they could add some heuristics to make it smarter.

Thanks for the patch!

Fuco1:
On second thought... I would like to have the elisp there to be copy-pastable from the docstring even without looking at it through the help(ful) buffer.  Doesn't make sense for the code in the functional file but the examples in dash file are at least syntactically valid (and can be quickly adapted).  Pasting something with `\\=` in it is going to cause trouble.

But I might be looking too deep into it.  I don't know how many people will find this annoying except me.

What do you think?

Wilfred:
> If people want curly quotes why not put curly quotes in the string?

I agree! Unfortunately, `text-quoting-style` defaults to `nil`,  so `'` is converted to curly quotes.

> I would like to have the elisp there to be copy-pastable from the docstring even without looking at it through the help(ful) buffer.

I agree here too. It's noisy and distracting. If people spend more time reading the source than using Help, then `\\=` is an inconvenience.

The bug this PR prevents is really hard to spot though. Here's the docstring from `-table-flat` with a curly quote:

```
(-flatten-n (1- (length lists)) (apply ’-table fn lists))
```

As far as I can tell, `\\=` is the proper Emacs approach. You can see it in built-ins, such as `apply` or `LaTeX-mode`:
```
Call FUNCTION with our remaining args, using our last arg as list of args.
Then return the value FUNCTION returns.
Thus, (apply \\='+ 1 2 \\='(3 4)) returns 10.
usage: (apply FUNCTION &rest ARGUMENTS)
```

In my current instance, I can see 177 functions that use escaping:

```emacs-lisp
(let (syms-with-escapes) 
  (mapatoms
   (lambda (s)
     (when (and
            (fboundp s)
            (ignore-errors (documentation s t))
            (s-contains-p "\\=" (documentation s t)))
       (push s syms-with-escapes))))
  (length syms-with-escapes))
```

Fuco1:
Yea, I guess that's not much to discuss.  I'm not against the PR it just really rubs me the wrong way :/

Won't this mess with the documentation we autogenerate?  We will probably have to strip that `\\='` and replace it with just `'`.

Wilfred:
Fortunately, documentation is still generated correctly. This is because doing `(documentation 'some-function)` automatically strips the escaping appropriately.

Issue #260, opened 2018-01-15
Remove prefix function


alphapapa:
I'm sure this is ugly and that there's a better way, but anyway: I needed a function to remove a prefix from a list.  This is not the same as `seq-difference` or `-difference`, because those operate on individual elements, not lists within lists.  I came up with this:
```el
(defun -remove-prefix (remove from)
  (if (-is-prefix? remove from)
      (let* ((prefixes (nreverse (-inits remove)))
             (longest-prefix (cl-loop collect (pop prefixes) into tested
                                      until (not (-is-prefix? (last tested) from))
                                      finally return (last tested))))
        (-slice from (1+ (length longest-prefix))))
    from))
```
Which works like:
```el
(-remove-prefix '("1" "John") '("1" "John" "1" "1"))   ;; => ("1" "1")
```
I'm guessing a function that does this might be a good addition.

Fuco1:
Do you want to remove the longest common prefix or only if they exactly match?

basil-conto:
If the outer `if` succeeds, i.e. if `remove` is already a prefix of `from`, then why is the calculation of `longest-prefix` necessary? Isn't `remove` already the longest prefix, since it satisfies `(-is-prefix-p remove from`)?

In other words, your current example is equivalent to

```el
(defun -remove-prefix (remove from)
   (if (-is-prefix-p remove from)
       (-slice from (length remove))
     from))
```

or

```el
(defun -remove-prefix (remove from)
  (--drop-while (and remove (equal it (pop remove)))
                from))
```

So, to rephrase https://github.com/magnars/dash.el/issues/260#issuecomment-357724497, are you interested in removing (a) `remove` only if it is already a prefix of `from`, or (b) the longest common prefix of `remove` and `from` from `from`?

Either way, my opinion is that the only useful function to add to `dash.el` would be something called `-prefix`, `-shared-prefix` or similar, which calculates the longest common prefix of a list of lists, unless this already exists?

This opinion is based on the fact that (a) is pretty trivial to implement using existing functions, as I have shown, and (b) is made trivial by a `-shared-prefix` function.

alphapapa:
You're right, `remove` is already the longest if `-is-prefix?` is non-nil.  I think I was too deep, not seeing the forest for the trees there.  Thanks.

basil-conto:
@alphapapa @Fuco1 WDYT of the suggestion to add a longest-common-prefix function? I'm not aware of any such function existing.

[...]

PR #261, opened 2018-01-16
Add -common-prefix

basil-conto:
Re: #260 

Using `-reduce` affords a very simple depth-first implementation, albeit one quite amenable to optimisation. In particular, it would be much better to take a breadth-first approach, but this should suffice for an initial version, right?


Fuco1:
That's a pretty cool implementation :) Thanks!

PR #262, opened 2018-01-17
Safer -common-prefix via breadth-first traversal

basil-conto:
This implementation is slightly slower than using `--reduce` under "normal" circumstances, but correctly short-circuits performance pitfalls such as

```el
(let ((l (number-sequence 0 1000)))
  (-common-prefix l l ()))
```

1. Do you think it's worth complicating the implementation to protect against such scenarios?
2. Should I annotate the various local variables used with a brief description of their purpose, or do you find this clear enough?

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -2118,10 +2118,7 @@ or with `-compare-fn' if that's non-nil."
 (defun -common-prefix (&rest lists)
   "Return the longest common prefix of LISTS."
   (declare (pure t) (side-effect-free t))
-  (--reduce (let (head prefix)
-              (while (and acc it (equal (setq head (pop acc)) (pop it)))
-                (push head prefix))
-              (nreverse prefix))
+  (--reduce (--take-while (and acc (equal (pop acc) it)) it)
             lists))
 
 (defun -contains? (list element)


Fuco1:
I was thinking of doing something like this:

1. initialize `prefixLength` to 0
2. loop over all the input lists
  1. take the first item from each list, check if `-uniq` of them is of length 1 (can be O(n) scan instead of uniq which is quadratic)
  2. if yes, increment `prefixLength`
  3. if not, endloop
3. return `-take prefixLength (car lists)`

Whould this not work?  There's a lot of setq and pops in your code which I find a bit confusing.  It probably works the same way as I describe but it isn't obvious to me at a glance.

This is the implementation I came up with in 3 minutes doing the above algorithm.  Not the most optimal, maybe we are in fact doing the same thing :) But just to illustrate

```elisp
(defun -common-prefix (&rest lists)
  (let ((prefixes (-map (lambda (x)
                          (if (and (consp x)
                                   (not (listp (cdr x))))
                              (list (car x) (cdr x))
                            x))
                        (-unzip lists)))
        (prefix-length 0))
    (while (= 1 (length (-uniq (pop prefixes)))) (cl-incf prefix-length))
    (-take prefix-length (car lists))))


(-common-prefix '(1 2 3) '(1 2) '(1 2 3 4))
```

basil-conto:
### Summary

> There's a lot of `setq` and `pops` in your code which I find a bit confusing.

Yes, I went a bit crazy with imperative++ :)

> It probably works the same way as I describe but it isn't obvious to me at a glance.

Our ideas only superficially seem similar because they follow the same basic notion&mdash;that of checking each "column" (assuming row-major order) of the input "matrix" at a time. The way we approach this, however, is quite distinct in terms of algorithmic complexity.

In a nutshell, the algorithm proposed in this PR is the following:

- IF matrix has >= 2 rows, THEN
  + initialise prefix to nil
  + WHILE no rows are empty AND all heads are equal
    - push head onto prefix
  + return reversed prefix
- ELSE return first row

The tricky part is doing this linearly for each column without inefficiently transposing the matrix upfront via `-unzip` or similar. This is why I use a secondary variable called `tails` in my implementation, which accumulates the section of the matrix to be used in the next iteration. In other words, `tails` is a list of the `cdr`s of the input `lists`, which is updated for each outer loop.

### Details

#### Dramatis personae

| Variables | Description                                                  |
|-----------|--------------------------------------------------------------|
| `x`:`xs`  | first row in matrix                                          |
| `y`:`ys`  | assigned to each subsequent row in turn                      |
| `tails`   | copy of matrix with all row heads removed for next iteration |
| `prefix`  | result accumulator                                           |
| `eq`      | whether all column elements were equal                       |

#### Run-through

- LOOP
  + BREAK if first row is empty
  + `x`:`xs` <- pop first row from matrix
  + LOOP
    - BREAK if first row is empty
    - `y`:`ys` <- pop first row from matrix
    - BREAK if `y` &ne; `x`
    - push `ys` onto `tails`
  + BREAK if there was some `y` &ne; `x`
  + push `x` onto `prefix`
  + push `xs` onto `tails`
  + matrix <- `tails`
- return reversed `prefix`

### Comparison

> Whould this not work?

It would, although it is, as you say, not optimal. Here is your suggestion with a couple of obvious bug fixes and optimisations:

```el
(defun prefix--unzip-take (&rest lists)
  (let ((prefixes (if (and (cdr lists) (not (cddr lists)))
                      (mapcar #'-cons-to-list (-unzip lists))
                    (-unzip lists)))
        (prefix-length 0))
    (while (and prefixes (null (cdr (-uniq (pop prefixes)))))
      (setq prefix-length (1+ prefix-length)))
    (-take prefix-length (car lists))))
```

Here is a variant of this which avoids calls to `-uniq` and `-take`:

```el
(defun prefix--unzip-accumulate (&rest lists)
  (let ((cols (if (and (cdr lists) (not (cddr lists)))
                  (mapcar #'-cons-to-list (-unzip lists))
                (-unzip lists)))
        col head prefix)
    (while (when cols
             (setq col (pop cols))
             (setq head (pop col))
             (--all-p (equal it head) col))
      (push head prefix))
    (nreverse prefix)))
```

Finally, here are renamed listings of my incumbent and proposed implementations:

```el
(defun prefix--reduce (&rest lists)
  (--reduce (let (head prefix)
              (while (and acc it (equal (setq head (pop acc)) (pop it)))
                (push head prefix))
              (nreverse prefix))
            lists))

(defun prefix--linear (&rest lists)
  (if (and lists (cdr lists))
      (let (x xs ys tails prefix eq)
        (while (when (setq xs (pop lists))
                 (setq x (pop xs))
                 (while (when (setq eq (and (setq ys (pop lists))
                                            (equal x (pop ys))))
                          (push ys tails)
                          lists))
                 eq)
          (push x prefix)
          (setq lists (cons xs tails))
          (setq tails ()))
        (nreverse prefix))
    (car lists)))
```

First, some peace of mind:

```el
(setq l (number-sequence 0 1000))
(setq fns (list #'prefix--unzip-take
                #'prefix--unzip-accumulate
                #'prefix--reduce
                #'prefix--linear))

;; Check all implementations give the same result
(--all-p (= 1 (length (-uniq (mapcar (-cut apply <> it) fns))))
         `(()
           (())
           (() ())
           ((1))
           ((1) ())
           (() (1))
           ((1 2) ())
           (() (1 2))
           ((1 2) () (1 2))
           (() (1 2) (1 2))
           ((1 2) (1 2) ())
           ((1 2) (3 4))
           ((1 2 3 4) (1 2 3 4 5))
           (,l ,l ,l)
           (,l ,l ())))
  ;; ⇒ t
```

Then, some amortised profiling:

```el
(mapc #'byte-compile fns)

(benchmark-run-compiled 1000
  (prefix--unzip-take)
  (prefix--unzip-take ())
  (prefix--unzip-take () ())
  (prefix--unzip-take '(1))
  (prefix--unzip-take '(1) ())
  (prefix--unzip-take () '(1))
  (prefix--unzip-take '(1 2) ())
  (prefix--unzip-take () '(1 2))
  (prefix--unzip-take '(1 2) () '(1 2))
  (prefix--unzip-take () '(1 2) '(1 2))
  (prefix--unzip-take '(1 2) '(1 2) ())
  (prefix--unzip-take '(1 2) '(3 4) '(1 2) '(3 4) '(1 2) '(3 4) '(1 2) '(3 4) '(1 2) '(3 4) '(1 2) '(3 4) '(1 2) '(3 4) '(1 2) '(3 4) '(1 2) '(3 4) '(1 2) '(3 4) '(1 2) '(3 4))
  (prefix--unzip-take '(1 2 3 4) '(1 2 3 4 5))
  (prefix--unzip-take '(1 2 3 4) '(1 2 3 4 5) '(1 2 3 4) '(1 2 3 4 5) '(1 2 3 4) '(1 2 3 4 5) '(1 2 3 4) '(1 2 3 4 5) '(1 2 3 4) '(1 2 3 4 5) '(1 2 3 4) '(1 2 3 4 5))
  (prefix--unzip-take l l l)
  (prefix--unzip-take l l l l l l)
  (prefix--unzip-take l l l l ()))
```

Repeating the same for all implementations results in the following rough averages on my machine:

| Function                   | CPU Time (s) | GC (#) |
|----------------------------|--------------|--------|
| `prefix--unzip-take`       | 8.6          | 83     |
| `prefix--unzip-accumulate` | 7.3          | 76     |
| `prefix--reduce`           | 2.9          | 34     |
| `prefix--linear`           | 3.3          | 37     |

### Conclusion

IMO:

1. It's not worth moving away from the incumbent implementation towards one which relies on `-unzip`.
2. If we want to avoid the depth-first performance pitfall of `prefix--reduce`, we should bite the bullet and use `prefix--linear`.

Given the data above, I'm happy sticking with `prefix--reduce`.

WDYT?

Fuco1:
I think you could do without `tails`, `pop` should update the rest in place no?

Yea, the unzip implementation was a bit of a simplification, a "pseudocode" if you will, :P But the ideas are quite similar in terms of the column scan.

So the contenders are reduce or linear version, not quite sure.  The reduce one is a bit simpler to understand so unless there's some extreme need for optimization I would probably go with the simpler implementation.    Good job on the benchmark so we know what we're sitting at!

There could maybe be some smarts to look at the inputs and if there are some really long ones and some really short ones pick another implementation... but then that is also going to slow things down and make it even more complex.

basil-conto:
> I think you could do without `tails`, `pop` should update the rest in place no?

I don't think this is possible while traversing the list:

```el
(let* ((lists '((1 2 3) (1 2) (1 2 3 4)))
       (rest  (car lists)))
  (pop rest)
  lists)
;; ⇒ ((1 2 3)
;;    (1 2)
;;    (1 2 3 4))
```

Keep in mind we need to maintain a reference to the original head of `lists` from which to start the outer loop. Unless I'm being obtuse and missing something obvious?

> Yea, the unzip implementation was a bit of a simplification, a "pseudocode" if you will, :P But the ideas are quite similar in terms of the column scan.

> So the contenders are reduce or linear version

Here's another contender, a compromise between the pseudocode unzip and unwieldy linear versions:

```el
(defun prefix--mapcar (&rest lists)
  (let (head prefix)
    (while (when (and lists (--all-p it lists))
             (setq head (caar lists))
             (--all-p (equal (car it) head) lists))
      (push head prefix)
      (setq lists (mapcar #'cdr lists)))
    (nreverse prefix)))
```

And here are the relevant amortised results using the bench from my last message:

| Function         | CPU Time (s) | GC (#) |
|------------------|--------------|--------|
| `prefix--mapcar` | 4.4          | 42     |
| `prefix--reduce` | 2.6          | 38     |
| `prefix--linear` | 3.1          | 42     |


> The reduce one is a bit simpler to understand so unless there's some extreme need for optimization I would probably go with the simpler implementation.

I'm a bit torn between the `mapcar` and `reduce` versions, but leaning towards the latter purely for aesthetic reasons. :) Someone has to make an executive decision. :)

> There could maybe be some smarts to look at the inputs and if there are some really long ones and some really short ones pick another implementation... but then that is also going to slow things down and make it even more complex.

Yeah, I doubt that would be worth it.

Fuco1:
Rarely we get it right the first time but here it seems to have worked out after all :D Let's leave it as it is then! Thanks for all the thoughts you put into this!

While at it, wouldn't you be interested in also adding `-common-suffix`? :D

basil-conto:
> Rarely we get it right the first time

Perhaps this time is no exception. :) Please see my new commit and let me know if I should open a new PR for this.

> While at it, wouldn't you be interested in also adding `-common-suffix`?

Sure, I'll have a stab at it later if no-one beats me to it. :)

Fuco1:
Oh take-while, very nice :)  This is actually a lot simpler to understand.  It's only a step towards a non-side-effecting version... but I don't think we can do that lazily in elisp (in haskell one would e.g. zip the lists but that would only go as far as necessary).

I like the change here, shall we merge?

basil-conto:
> It's only a step towards a non-side-effecting version... but I don't think we can do that lazily in elisp (in haskell one would e.g. zip the lists but that would only go as far as necessary).

Elisp generators come to mind, but I suspect that would add too much complexity, even if feasible.

> shall we merge?

If you approve, sure. :)

Fuco1:
Also generators are uber-slow in elisp.

PR #263, opened 2018-01-18
Add -common-suffix

basil-conto:
Update `-common-prefix` tests along with those of `-common-suffix`.

Re: #262 

Moral question: Is it worth adding this function when users bent on bad practice like this can write the code themselves?

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -2310,6 +2310,10 @@ or with `-compare-fn' if that's non-nil."
   (--reduce (--take-while (and acc (equal (pop acc) it)) it)
             lists))
 
+(defun -common-suffix (&rest lists)
+  "Return the longest common suffix of LISTS."
+  (nreverse (apply #'-common-prefix (mapcar #'reverse lists))))
+
 (defun -contains? (list element)
   "Return non-nil if LIST contains ELEMENT.
 
@@ -2904,6 +2908,7 @@ structure such as plist or alist."
                              "-inits"
                              "-tails"
                              "-common-prefix"
+                             "-common-suffix"
                              "-contains?"
                              "-contains-p"
                              "-same-items?"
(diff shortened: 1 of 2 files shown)


Fuco1:
What do you mean by bad practice?

There's quite a lot of wrappers like this in dash where we do the "double reverse" trick.  I think it is convenient to have it both ways.

basil-conto:
> What do you mean by bad practice?

By "bad" I mean inefficient and non-idiomatic. Because Elisp lists are singly linked, idiomatic programs are usually designed so as to avoid having to do this double reversal.

> There's quite a lot of wrappers like this in dash where we do the "double reverse" trick. I think it is convenient to have it both ways.

But I guess this is a library and it should provide such conveniences without complaining too much. :)

Fuco1:
One of the really cool thing about having a lot of combinators is that it helps with prototyping.  I can do quite a lot with very little code when the library provides enough building blocks (and obviously I know dash quite well at this point :D).  So while not ready for production code this is quite nice to have when trying something out.

basil-conto:
You're right, and I appreciate that. :)

By the way, I tried writing `-common-suffix` in terms of `--reduce-from` from scratch, rather than reusing `-common-prefix`, but this way is just as performant. Let me know if there's anything else you'd like me to iron out before merging this.

basil-conto:
P.S. Any idea what's causing the (unrelated) Emacs 23 infloop in the tests?

basil-conto:
Friendly ping.

Fuco1:
Friendly pong!  I've assigned this to myself so it will show up in my planning, hopefully I'll get to it.

basil-conto:
Ping, thanks. :)

basil-conto:
I removed the doc changes created by `create-docs.sh` to avoid merge conflicts, so if and when this gets merged they should be regenerated.

Anything else holding this up?

PR #265, opened 2018-02-07
Add missing keywords to dash-enable-font-lock

Wilfred:
Ensure that all the dash functions and macros are highlighted by
dash-enable-font-lock.

(It's possible I've missed a few, but I diffed the output from `imenu` against this list and added all the missing ones shown.)


magnars:
Thanks! 👍

PR #266, opened 2018-03-06
Use EVM for travis, test recent Emacs

phillord:
This switches Emacs installation for travis from PPAs to EVM. This
also allows testing of more recent version (Emacs-25 prerelease and
snapshot.

Currently, the tests are failing for Emacs-23, although they have been failing on master for a while anyway. I did a slight amount of digging -- the most immediate test failure (if-let*) can be fixed by bumping up max-eval-depth, but then we get other failures. So, question -- should dash.el support Emacs-23 still? Is it worth fixing the bugs (hard but not impossible?) Or just disable the failing tests on Emacs-23 and hope that the rest works? Happy to attempt any of these.


magnars:
Thanks for working on this, Phil. 👏 

I think it's time to disable the tests for E23.

Fuco1:
There's already https://github.com/magnars/dash.el/tree/feature/coverall-integration which also tries to add coverage info but somehow I failed to get it working.

phillord:
@Fuco1 Okay -- so what you want to do? I can remove Emacs-23 from this patch. Or is your branch doing the same thing wrt EVM?

Fuco1:
@phillord I think we can close mine because the coverage reporting is still rather tricky to get working... however I've done some cleanup of the scripts which are really not necessary anymore so you might want to get some inspiration from that (e.g. run-travis is useless if we use cask/evm)

The pull request is here: https://github.com/magnars/dash.el/pull/195/files

Fuco1:
@phillord Yea it also switches to EVM.  Also I would prefer to test all the minor versions rather then just the last one.  There was quite a bit of changes from 24.2 to 24.3 for example (John said himself that he would've made it a major release had he been the maintainer back then)

phillord:
@Fuco1 It was 24.3->24.4 actually (nadvice.el being the main thing). Trivial to put those in.

We *don't* use cask currently, actually, on travis, because it fails for Emacs-23.

phillord:
Okay, think I have got most of it now. Emacs 23 support has gone. We can't remove the "emacs23" support in dash.el, because it's actually need for Emacs24 before gv.el. I've updated run-tests.el to be a skip emacs-24.1/24.2 byte-compile check, because I cannot work out how to pacify the byte-compiler for these versions. run-travis-ci I've left in place because it prints out nice stuff about the version which is useful I think. Oh, and ert.el is gone which makes life cleaner I think.

Any thing I have missed? If not I will squash and write a nice commit message.

Fuco1:
Yea you're right.  I opted for cask for the undercover dependency and it makes it a bit easier running emacs with proper load-path etc, but that is quite trivial as dash has no external dependencies.

I'm OK with the changes here, LGTM :+1:

phillord:
All squashed, and ready to go.

magnars:
I added you as a collaborator. You do the honors, sir. :-)

PR #269, opened 2018-07-11
Smart kv destructuring

### LINKED ISSUE #111: Smarter &plist, &alist destructuring
Most of the time we write the destructure like this:

```
(&plist :key key :otherkey otherkey)
```

where the variable to which we destructure is the same as they key sans the `:`. I'm proposing this: when `*` is appended to the selector (so it's e.g. `&plist*`) the name of the binding variable is derived from the key in the following "smart way":
- a key `:foo` is converted into `foo` variable
- a key `foo` is converted into `foo` variable
- a key `"foo"` is converted into `foo` variable

This covers pretty much 99% of use-cases of mapping structures.

Ideas?  Instead of `*` we might use a different symbol, but stars often means some variation in lisp so I think it fits well.

Fuco1:
Fixes #111 

- [x] Implement the feature
- [x] Update documentation
- [x] write tests, lots and lots of tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -2324,14 +2324,17 @@ Key/value stores:
     (&plist key0 a0 ... keyN aN) - bind value mapped by keyK in the
                                    `source` plist to aK.  If the
                                    value is not found, aK is nil.
+                                   Uses `plist-get` to fetch values.
 
     (&alist key0 a0 ... keyN aN) - bind value mapped by keyK in the
                                    `source` alist to aK.  If the
                                    value is not found, aK is nil.
+                                   Uses `assoc` to fetch values.
 
     (&hash key0 a0 ... keyN aN) - bind value mapped by keyK in the
                                   `source` hash table to aK.  If the
                                   value is not found, aK is nil.
+                                  Uses `gethash` to fetch values.
 
 Further, special keyword &keys supports "inline" matching of
 plist-like key-value pairs, similarly to &keys keyword of
@@ -2341,6 +2344,36 @@ plist-like key-value pairs, similarly to &keys keyword of
 
 This binds `n` values from the list to a1 ... aN, then interprets
 the cdr as a plist (see key/value matching above).
+
+`a` shorthand notation for kv-destructuring exists which allows the
+patterns be optionally left out and derived from the key name in
+the following fashion:
+
+- a key :foo is converted into `foo` pattern,
+- a key 'bar is converted into `bar` pattern,
+- a key "baz" is converted into `baz` pattern.
+
+That is, the entire value under the key is bound to the derived
+variable without any further destructuring.
+
+This is possible only when the form following the key is not a
+valid pattern (i.e. not a symbol, a cons cell or a vector).
+Otherwise the matching proceeds as usual and in case of an
+invalid spec fails with an error.
+
+Thus the patterns are normalized as follows:
+
+     ;; derive all the missing patterns
+     (&plist :foo 'bar "baz") => (&plist :foo foo 'bar bar "baz" baz)
+
+     ;; we can specify some but not others
+     (&plist :foo 'bar explicit-bar) => (&plist :foo foo 'bar explicit-bar)
+
+     ;; nothing happens, we store :foo in x
+     (&plist :foo x) => (&plist :foo x)
+
+     ;; nothing happens, we match recursively
+     (&plist :foo (a b c)) => (&plist :foo (a b c))
 
 You can name the source using the syntax `symbol` &as `pattern`.
 This syntax works with lists (proper or improper), vectors and
--- a/dash.el
+++ b/dash.el
@@ -1618,7 +1618,7 @@ SOURCE is a proper or improper list."
         (cond
          ((and (symbolp (car match-form))
                (memq (car match-form) '(&keys &plist &alist &hash)))
-          (dash--match-kv match-form (dash--match-cons-get-cdr skip-cdr source)))
+          (dash--match-kv (dash--match-kv-normalize-match-form match-form) (dash--match-cons-get-cdr skip-cdr source)))
          ((dash--match-ignore-place-p (car match-form))
           (dash--match-cons-1 (cdr match-form) source
                               (plist-put props :skip-cdr (1+ skip-cdr))))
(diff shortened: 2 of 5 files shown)


dash.el
```diff
@@ -1618,7 +1618,7 @@ SOURCE is a proper or improper list."
         (cond
          ((and (symbolp (car match-form))
                (memq (car match-form) '(&keys &plist &alist &hash)))
-          (dash--match-kv match-form (dash--match-cons-get-cdr skip-cdr source)))
+          (dash--match-kv (dash--match-kv-normalize-form match-form) (dash--match-cons-get-cdr skip-cdr source)))
```
Fuco1: Rename the function to `...-normalize-match-form`.

dash.el
```diff
@@ -1702,6 +1702,72 @@ is discarded."
...
+
+- a key :foo is converted into `foo' match-form,
+- a key 'bar is converted into `bar' match-form,
+- a key \"baz\" is converted into `baz' match-form.
+
+That is, the entire value under the key is bound to the derived
+variable without any further destructuring.
+
+This is possible only when the form following the key is not a
+valid match-form (i.e. a keyword, quoted symbol or a string).
```
Fuco1: describe valid match forms instead of an incomplete example of invalid match forms.

dash.el
```diff
@@ -1702,6 +1702,72 @@ is discarded."
...
+Otherwise the matching proceeds as usual and in case of an
+invalid spec fails with an error.
+
+This method normalizes such a match-form to the format expected
+by `dash--match-kv'."
+  (let ((normalized (list (car match-form)))
+        (skip nil))
+    (-each (apply '-zip (-pad '&--dash--fill-placeholder (cdr match-form) (cddr match-form)))
```
Fuco1: I guess we should generate the filler placeholder instead, just to be sure

dash.texi
```diff
@@ -3582,6 +3585,36 @@ plist-like key-value pairs, similarly to &keys keyword of
 This binds @var{n} values from the list to a1 ... aN, then interprets
 the cdr as a plist (see key/value matching above).
 
+@var{a} shorthand notation for kv-destructuring exists which allows the
+patterns be optionally left out and derived from the key name in
```
alphapapa: "patterns to be"  :)

dash.info
```diff
@@ -2262,6 +2265,36 @@ control.
      This binds N values from the list to a1 ...  aN, then interprets
      the cdr as a plist (see key/value matching above).
 
+     A shorthand notation for kv-destructuring exists which allows the
+     patterns be optionally left out and derived from the key name in
```
alphapapa: "patterns to be"  :)

dash.el
```diff
@@ -1894,6 +1938,36 @@ plist-like key-value pairs, similarly to &keys keyword of
 This binds N values from the list to a1 ... aN, then interprets
 the cdr as a plist (see key/value matching above).
 
+A shorthand notation for kv-destructuring exists which allows the
+patterns be optionally left out and derived from the key name in
```
alphapapa: "patterns to be"  :)

README.md
```diff
@@ -2342,6 +2345,36 @@ plist-like key-value pairs, similarly to &keys keyword of
 This binds `n` values from the list to a1 ... aN, then interprets
 the cdr as a plist (see key/value matching above).
 
+`a` shorthand notation for kv-destructuring exists which allows the
+patterns be optionally left out and derived from the key name in
```
alphapapa: "patterns to be"  :)

README.md
```diff
@@ -2342,6 +2345,36 @@ plist-like key-value pairs, similarly to &keys keyword of
 This binds `n` values from the list to a1 ... aN, then interprets
 the cdr as a plist (see key/value matching above).
 
+`a` shorthand notation for kv-destructuring exists which allows the
```
alphapapa: I think the first letter here shouldn't be in backticks.

alphapapa:
Thanks, Matus!  I've been wanting something like this in elisp for a long time.  I feel like this deserves a blog post, if you are so inclined, so that Emacs devs will know about this new ability.  :)

README.md
Fuco1: Yea, I've noticed this too.  The problem is that the sentence starts with `A` which the script thinks is a variable :D And so fontifies it accordingly.  I think I'll need to reorder the sentence to start with a regular word.

dash.texi
alphapapa: I'm not an expert on the English subjunctive, but I think a problem here is that you use "exists" followed by "allows," which at least implies the indicative mood.  e.g. this from the Wikipedia article (I hate to cite Wikipedia, but a few minutes of googling didn't find a better source yet):

> The present subjunctive is identical to the bare infinitive (and imperative) of the verb in all forms. This means that, for almost all verbs, the present subjunctive differs from the present indicative only in the third person singular form, which lacks the ending -s in the subjunctive.

I'm sure that one could find archaic "English" uses of nearly anything imaginable (who makes the rules, anyway?), but as a native speaker, "which allows the patterns be" feels ungrammatical at worst, and very awkward at best.  :)

alphapapa: Also, "optionally" seems redundant.

dash.el
```diff
@@ -1894,6 +1938,36 @@ plist-like key-value pairs, similarly to &keys keyword of
...
+the following fashion:
+
+- a key :foo is converted into `foo' pattern,
+- a key 'bar is converted into `bar' pattern,
+- a key \"baz\" is converted into `baz' pattern.
+
+That is, the entire value under the key is bound to the derived
```
alphapapa: IMHO "under the key" is a bit awkward.  Perhaps this would be clearer:

> That is, the entire value mapped to by the key...

dash.texi
Fuco1: The subjunctive clause is `allows be` not `exists allows`, but I agree and have no problem to change it.

Fuco1: Also for some reason I had the impression you were german... I guess it's the github picture, it looks  very similar to one german user I know :D Oh the silly human brain.

dash.el
Fuco1: Maybe you could do a PR :D Feel free to improve the docs, they sure are sloppy at times.  Don't forget to run `./create-docs.sh`!

dash.texi
alphapapa: Haha, no, I'm American.

dash.el
alphapapa: > Feel free to improve the docs, they sure are sloppy at times. 

No promises, but I'll keep that in mind.

basil-conto: FWIW, I've never heard of a value existing under a key, and I wouldn't say "the entire value mapped to by the key", as this is slightly ambiguous (keys don't actively map to values, rather they are mapped to values by the containing data structure).

I think for Elisp it's more natural to say things like "the value associated with the key", "the value corresponding to the key", etc.

(Disclaimer: I'm not a native English speaker :)

PR #272, opened 2018-07-26
Implement -dolist

### LINKED ISSUE #136: FR: Add variant of -each with -let binding
Right now, we need

```elisp
(-each '((a . b) (c . d))
  (-lambda ((first . second)) (code-using first second))
```

Could be

```elisp
(-??? (first . second) '((a . b) (c . d))
  (code-using first second))
```

Similar like `cl-dolist` with destructuring.

Fuco1:
Fixes #136 

The syntax is as follows

```elisp
(-dolist ((first . second) '((a . b) (c . d)))
  (message "first %s second %s" first second))

;; also works as usual `dolist'
(-dolist (var '((a . b) (c . d)))
  (message "first %s" var))
```

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -92,6 +92,22 @@ the target form."
 
 (put '-each 'lisp-indent-function 1)
 
+(defmacro -dolist (varlist &rest body)
+  "Loop over a list.
+
+Evaluate BODY with MATCH-FORM bound to each car from LIST, in turn.
+
+MATCH-FORM destructuring is done according to the rules of `-let'.
+
+See also: `dolist', `-each'.
+
+\(fn (MATCH-FORM LIST) BODY...)"
+  (declare (indent 1))
+  (let ((source (make-symbol "--dash-source--")))
+    `(dolist (,source ,(cadr varlist))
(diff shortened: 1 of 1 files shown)


Wilfred:
How about the name `-each-let`? That fits quite nicely with `-when-let` etc.

Wilfred:
As an aside, `dolist` is awesome when you want nested loops.

```emacs-lisp
(dolist (team world-cup)
  (dolist (player team)
    (foo player)))
```

Compare this with:

```emacs-lisp
(--each world-cup
  (--each it
    (foo it)))
```

This is confusing because `it` means different things on line 2 vs line 3. There's also no way to refer back to the outer `it` (which I've named `team` in the first example).

I don't like that `cl` advises `dolist`, and I think `dolist`'s optional result argument is less readable, but the basic use case is really handy.

Fuco1:
Yea, so we can view `dolist` is like an `--each` with named `it`.  Which I think `-each-let` represents nicely.  I think I'll go with that, thanks!

wbolster:
`--each-as` anyone? (cf. `-as->`)

(just thinking out loud)

Issue #276, opened 2018-11-19
Ability to define custom destructorings


yyoncho:
At this point, `dash` supports 4 different destructoring keyword (keys/hash/plist/alist) and from what I can see it will be not hard to add new keywords by providing a destructuring template(or macro?) to generate the getter.

Example usecases:

* `&hash` which checks whether the map is empty before applying gethash
* Custom destructuring for `cl-defstruct`, e. g. 
``` elisp
(-let [(&my/struct slot1 slot2) value]
  ...)
```

Fuco1:
Hi Ivan.  This is something I haven't considered but it shouldn't be terribly difficult as you say.  One would need to make sure the recursive expansions would still work, but so long as you do not publish your custom getter I guess the contract is your own.

The place to start at would be in the function `dash--match` in the second condition `(consp match-form)`.  Before the default case we can add something that would look at the symbol's property (using `get`) `dash--matcher` or something similar and execute that.

Would you be interested in trying to implement this (does what I said make any sense at all? :))

yyoncho:
I have checked the code upfront, just wanted to know whether you are going to accept such PR - check #277 . I have not updated the docs if this is needed.

yyoncho:
@Fuco1 is there a change to move this issue forward? In lsp-mode we are want to support both maps and alists and we are going to replace gethash with map-elt but we still do not have a solution for `&hash` destructoring. We spent a lot of time to migrate our code to use dash and it will be a pity if we now have to rewrite it again to use `pcase-let` or something like that.

Fuco1:
@yyoncho I will merge this today, I promise!

yyoncho:
Thank you @Fuco1 . Not sure whether it is ready for merging though (e. g. it lacks documentation).

Issue #179, opened 2016-05-01
Discussion: should we merge with seq.el?


Fuco1:
`seq` is now in emacs core and available to all, a move which we somehow failed to push for with dash.

It still only supports a rather small subset of `dash` functionaity, which is a reason I haven't adopted it yet (and likely never will until it is feature-compatible).

The move would require some effort but would solidify the emacs API, move the less-often-needed-but-really-nice-to-have functions into emacs proper, remove one dependency for great many packages, add support for multiple sequence types (granted we would have to reimplement some functions to work universally).  Plus having one API instead of two is reason in itself I think.  One problem that stands out to me immediately is how to deal with anaphora.

Magnar and I poured a lot of time into this project, but maybe it is time to integrate it and then phase it out eventually.

I would like to CC some people who I know depend (or did in the past, moving to seq in the meanwhile) and hear your opinions: @tarsius @rejeep @lunaryorn @bbatsov @nicferrier.  If you know someone else who you think should have a say please invite them too.

Fuco1:
Oh, and of course @NicolasPetton :D

bbatsov:
I'm all for a "merger" of the projects. I believe that packages like `seq` and `map` are definitely the way to go (nothing builds less 3rd party deps), plus they are really clean and have great focused APIs. I've started moving away from `dash.el` in the projects I maintain a while back. I think many of the useful things that don't belong to seq and map can simply be stuffed into subr-x.el (I'd place there some of the s.el functions as well).

Frankly I don't think the anaphoras are that important, as they produce slightly confusing code for people not familiar with them. I used them a lot initially and then I simply stopped (even while I was still on dash). 

//cc @malabarba

magnars:
I think `seq.el` is a great initiative from @NicolasPetton. Dash was never going to be approved by emacs-devel, due to its cheeky namespace and weird anaphoras. So in my mind, `seq` is already the "dash in core" initiative. I quite like the anaphoras, but like @bbatsov I don't think they're very important. I also think they're unlikely to be included in core.

I think the best way forward for Emacs is to expand `seq` to include more of the dash features, moving towards phasing dash out. And in the meantime, dash will continue to be a bunch of helpful features that should have been in core but aren't there yet.

Malabarba:
So far, whenever I proposed some sequence-related functionality to Nico, he was happy to accept it into seq. 

[...]

PR #277, opened 2018-11-19
Provided option to extend destructoring

yyoncho:
The descructoring could be extended by defining a function named
`dash-expand:&{destructoring-key}` where {destructoring-key} stands for the key
that will be used in the let destructoring, e. g. &hash/&plist in the standart
dash

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1650,6 +1650,10 @@ All returned symbols are guaranteed to be unique."
      (t
       (cons (list s source) (dash--match-cons-1 match-form s))))))
 
+(defun dash--get-expand-function (type)
+  "Get expand function name for TYPE."
+  (intern (format "dash-expand:%s" type)))
+
 (defun dash--match-cons-1 (match-form source &optional props)
   "Match MATCH-FORM against SOURCE.
 
@@ -1669,7 +1673,7 @@ SOURCE is a proper or improper list."
        ((cdr match-form)
         (cond
          ((and (symbolp (car match-form))
-               (memq (car match-form) '(&keys &plist &alist &hash)))
+               (functionp (dash--get-expand-function (car match-form))))
           (dash--match-kv (dash--match-kv-normalize-match-form match-form) (dash--match-cons-get-cdr skip-cdr source)))
          ((dash--match-ignore-place-p (car match-form))
           (dash--match-cons-1 (cdr match-form) source
@@ -1810,6 +1814,25 @@ kv can be any key-value store, such as plist, alist or hash-table."
      (t
       (cons (list s source) (dash--match-kv-1 (cdr match-form) s (car match-form)))))))
 
+(defun dash-expand:&hash (key source)
+  "Generate extracting KEY from SOURCE for &hash destructuring."
+  `(gethash ,key ,source))
+
+(defun dash-expand:&plist (key source)
+  "Generate extracting KEY from SOURCE for &plist destructuring."
+  `(plist-get ,source ,key))
+
+(defun dash-expand:&alist (key source)
+  "Generate extracting KEY from SOURCE for &alist destructuring."
+  `(cdr (assoc ,key ,source)))
+
+(defun dash-expand:&hash? (key source)
+  "Generate extracting KEY from SOURCE for &hash? destructuring.
+Similar to &hash but check whether the map is not nil."
+  `(when ,source (gethash ,key ,source)))
+
+(defalias 'dash-expand:&keys 'dash-expand:&plist)
+
 (defun dash--match-kv-1 (match-form source type)
   "Match MATCH-FORM against SOURCE of type TYPE.
 
@@ -1827,13 +1850,8 @@ Valid values are &plist, &alist and &hash."
                  (lambda (kv)
                    (let* ((k (car kv))
                           (v (cadr kv))
-                          (getter (cond
-                                   ((or (eq type '&plist) (eq type '&keys))
-                                    `(plist-get ,source ,k))
-                                   ((eq type '&alist)
-                                    `(cdr (assoc ,k ,source)))
-                                   ((eq type '&hash)
-                                    `(gethash ,k ,source)))))
+                          (getter
+                           (funcall (dash--get-expand-function type) k source)))
                      (cond
                       ((symbolp v)
                        (list (list v getter)))
@@ -1866,7 +1884,7 @@ Key-value stores are disambiguated by placing a token &plist,
       (let ((s (car match-form)))
         (cons (list s source)
               (dash--match (cddr match-form) s))))
-     ((memq (car match-form) '(&keys &plist &alist &hash))
+     ((functionp (dash--get-expand-function (car match-form)))
       (dash--match-kv (dash--match-kv-normalize-match-form match-form) source))
      (t (dash--match-cons match-form source))))
    ((vectorp match-form)
--- a/dev/examples.el
+++ b/dev/examples.el
@@ -28,6 +28,11 @@
 (defun square (num) (* num num))
 (defun three-letters () '("A" "B" "C"))
 
+(defun dash-expand:&hash-or-plist (key source)
+  "Sample destructoring which works with plists and hash-tables."
+  `(if (hash-table-p ,source) (gethash ,key ,source)
+     (plist-get ,source ,key)))
+
 ;; Allow approximate comparison of floating-point results, to work
 ;; around differences in implementation between systems. Use the `~>'
 ;; symbol instead of `=>' to test the expected and actual values with
@@ -1142,7 +1147,8 @@ new list."
       (puthash :foo 1 hash)
       (puthash :bar 2 hash)
       (-let (((&hash :foo :bar) hash)) (list foo bar))) => '(1 2)
-      (-let (((_ &keys :foo :bar) (list 'ignored :foo 1 :bar 2))) (list foo bar)) => '(1 2)
+    (-let (((&hash :foo (&hash? :bar)) (make-hash-table)))) => nil
+    (-let (((_ &keys :foo :bar) (list 'ignored :foo 1 :bar 2))) (list foo bar)) => '(1 2)
     ;;; go over all the variations of match-form derivation
     (-let (((&plist :foo foo :bar) (list :foo 1 :bar 2))) (list foo bar)) => '(1 2)
     (-let (((&plist :foo foo :bar bar) (list :foo 1 :bar 2))) (list foo bar)) => '(1 2)
@@ -1181,7 +1187,12 @@ new list."
     (-let [(list &as _ _ _ a _ _ _ b _ _ _ c) (list 1 2 3 4 5 6 7 8 9 10 11 12)] (list a b c list)) => '(4 8 12 (1 2 3 4 5 6 7 8 9 10 11 12))
     (-let (((x &as a b) (list 1 2))
            ((y &as c d) (list 3 4)))
-      (list a b c d x y)) => '(1 2 3 4 (1 2) (3 4)))
+      (list a b c d x y)) => '(1 2 3 4 (1 2) (3 4))
+    (-let (((&hash-or-plist :key) (--doto (make-hash-table)
+                                    (puthash :key "value" it))))
+      key) => "value"
+    (-let (((&hash-or-plist :key) '(:key "value")))
+      key) => "value")
 
   (defexamples -let*
     (-let* (((a . b) (cons 1 2))


dash.el
```diff
@@ -1661,7 +1661,7 @@ SOURCE is a proper or improper list."
        ((cdr match-form)
         (cond
          ((and (symbolp (car match-form))
-               (memq (car match-form) '(&keys &plist &alist &hash)))
+               (functionp (intern (format "dash-expand:%s" (car match-form)))))
           (dash--match-kv (dash--match-kv-normalize-match-form match-form) (dash--match-cons-get-cdr skip-cdr source)))
```
Fuco1: Does it make sense to call this `match-kv` now? That stands for key-value but clearly this custom thing can match anything.

There is also the call to `dash--match-kv-normalize-match-form` which for most other things makes no sense as it generates automatic places from key names.  If you have no keys then this will generate nonsense.

I think the check should go to a new branch.

yyoncho: I am not sure I understand the comment. Are you suggesting to rename match-kv related internal functions to something else?

As a side note, I found out that the elisp linter fails to handle the newly added destructorings. I am not sure how to fix that, I will try to make the extension functions `autoload`-able.

yyoncho: Ignore the last part, the problem with the checker what that it was using the `dash` that was downloaded from melpa, not the one that I have cloned.

Fuco1: I'm suggesting to create a new branch in the cond for the additional check and leave the kv stuff as it is because it has special handling for some kv related things which might not be applicable in general for any matcher that the user adds.

yyoncho: Are you are referring the `smarter` destructoring like `(&hash "key1" "key2")`? If this is the case IMO it would be better to keep it as in the PR in order to share that logic. If there is a need for destructoring for which the kv related stuff is not applicable there could be another extension point.

Fuco1: But this patch does not "extend kv" but anything.

Imagine you write a matcher for a list, writing `(&list a b)` would instead generate `(%list 'a a 'b b)` with the current code.  That makes no sense.

Fuco1: Or maybe it would just flat-out error because that can't in fact be expanded.

yyoncho: This PR targets only generation of KV templates -  it does not handle all of the cases and I haven't had that as a goal(although I cannot think of a pattern to which cannot be handled). I have already commented what are the benefits from sharing the kv related logic and you could see that in the added tests. If you do not like the PR and you think that does not bring value feel free to reject it. IMO it could be even viewed as a refactoring since it removed the &hash/&key/&alist/&plist usages through the code.

Furthermore, even your example `(&list a b)` will work if the KV logic is changed to use keys `0` and `1`...

Fuco1: I'm sorry but I don't understand.  You opened the issue https://github.com/magnars/dash.el/issues/276 and this is supposed to solve that.  But it does something else.  That's where my confusion is coming from.

Was the goal here not to address #276?

yyoncho: Here it is the description of the issue...

> At this point, `dash` supports 4 different destructoring keyword (keys/hash/plist/alist) and from what I can see it will be not hard to add new keywords by providing a destructuring template(or macro?) to generate the getter.
> 
> Example usecases:
> 
>     * `&hash` which checks whether the map is empty before applying gethash
> 
>     * Custom destructuring for `cl-defstruct`, e. g.
> 
> 
> ```emacs-lisp
> (-let [(&my/struct slot1 slot2) value]
>   ...)
> ```

... and I did exactly what I have described covering the usecases I have mentioned. Note that I have described the implementation details so the maintainers could comment upfront whether this approach is acceptable in order to avoid wasting my time doing a PR that wont be accepted.

Fuco1: The first part at no place mentions key/value destructuring, only that we have 4 keywords now (which incidentally manage kv storage).

Structures could be solved in some generic way out of the box instead of requiring the user to add a destructuring definition for every structure.

The `&hash?` example is interesting.  Maybe we could automatically extend every keyword with `?` to check for the condition somehow.

The point is that if we give the user ability to extend this in arbitrary ways we should make sure it works as expected.  I already misunderstood this feature and I'm quite familiar with the whole thing.  So I imagine other people can get confused as well.

I'll need to play with this a bit more.

I am not against adding support for custom expanders, but it needs to make sense.  I don't want to end up with another pcase.

basil-conto: > I am not against adding support for custom expanders, but it needs to make sense. I don't want to end up with another pcase.

FWIW, one person's `pcase` is another person's `pcase`. ;)

yyoncho: > The first part at no place mentions key/value destructuring, only that we have 4 keywords now (which incidentally manage kv storage).

It is not incidentally, but there is no point to discuss that...

> Structures could be solved in some generic way out of the box instead of requiring the user to add a destructuring definition for every structure.

Yes they can, but it would require the name of the structure to make that work: Compare:

``` emacs-lisp
(-let (((&lsp-wk 'proc (&lsp-cl 'server-id)) workspace)
   ...)
```
with...
``` emacs-lisp
(-let (((&slots lsp--workspace-parser proc (&slots lsp--client server-id)) workspace)
   ...)
```

>
> The `&hash?` example is interesting. Maybe we could automatically extend every keyword with `?` to check for the condition somehow.

No, since it makes sense only for &hash. &hash? could go as part of the `dash` core.

> The point is that if we give the user ability to extend this in arbitrary ways we should make sure it works as expected. I already misunderstood this feature and I'm quite familiar with the whole thing. So I imagine other people can get confused as well.

I am comfused even more by your comments - in previous comment you said that this destructuring is NOT powerful enough but now you wonder whether to give that power to the users...

> I'll need to play with this a bit more.
>
> I am not against adding support for custom expanders, but it needs to make sense. I don't want to end up with another pcase.

Sure, take your time. `pcase` issues could be avoided by having namespace prefix to each destructoring key.

yyoncho:
Ping

dash.el
```diff
@@ -1858,7 +1867,7 @@ Key-value stores are disambiguated by placing a token &plist,
       (let ((s (car match-form)))
         (cons (list s source)
               (dash--match (cddr match-form) s))))
-     ((memq (car match-form) '(&keys &plist &alist &hash))
+     ((functionp (intern (format "dash-expand:%s" (car match-form))))
```
Fuco1: @Fuco1 	Turn the `intern` logic into a helper function, reuse it also above in `kv-1`

dev/examples.el
```diff
@@ -28,6 +28,10 @@
 (defun square (num) (* num num))
 (defun three-letters () '("A" "B" "C"))
 
+(defun dash-expand:&hash? (key source)
```
Fuco1: @Fuco1 	This looks useful enough to be part of core, move it to `dash.el`

dash.el
```diff
@@ -1802,6 +1802,20 @@ kv can be any key-value store, such as plist, alist or hash-table."
...
 
+(defun dash-expand:&hash (key source)
+  "Generate extracting KEY from SOURCE for &hash destructuring."
+  `(gethash ,key ,source))
+
+(defun dash-expand:&plist (key source)
+  "Generate extracting KEY from SOURCE for &plist destructuring."
+  `(plist-get ,source ,key))
+
+(defun dash-expand:&alist (key source)
```
Fuco1: @Fuco1 	Add another destructuring key for structs

dash.el
yyoncho: Will do.

dev/examples.el
yyoncho: Sure.

dash.el
yyoncho: Can you elaborate?

dev/examples.el
yyoncho: In addition, I think that desctructoring that maps to `map-elt` might be usefull too(or dash reimplementation with similar behaviour). We are going to create such in lsp-mode but I believe it might be usefull for other dash clients.

dash.el
Fuco1: Oh, these were notes for myself!

I was thinking we could write a `&struct` to be a generic one and you could also write your own like you did in an example above for lsp.  But I think we should keep that out of this PR as it's not really related.

dev/examples.el
Fuco1: I'm not familiar with `map-elt`, is it some generic kv storage interface?  Does it make sense to use `&map` for it? (in other words, is it generic enough?)

Fuco1: Seems like pcase is using `map` matcher and there's `mapp` in core so let's use `&map`.

yyoncho: I am not very familiar too, the general idea is that it will `(map-elt m "key")` works with map/plist/alist. I think it is geneneric enough since you could extend it via defmethods to work with other structure(not 100% sure though). I will push another diff with &map implementation.

yyoncho: Actually, since `map` is new library it is not present in emacs < 25 do you want this in the core?

basil-conto: > I am not very familiar too, the general idea is that it will `(map-elt m "key")` works with map/plist/alist. I think it is geneneric enough since you could extend it via defmethods to work with other structure(not 100% sure though).

`map.el` was added to Emacs 25. It originally supported only alists, hash-tables, and arrays, and its API was not generic.

Emacs 27 reimplements the API as generic functions and adds support for plists. It also deprecates `map-put` and the `testfn` argument of `map-elt` and `map-contains-key`.

Depending on which version of Emacs you're targeting, it may be too early to use `map.el`. See also #179.

Fuco1: @yyoncho @basil-conto ok then we can maybe wait.  Or we can create another package `dash-map...` or something :D And put it there.  But that's outside of the scope of this patch.

I think all we need is to add documentation and it's good to go.

basil-conto: > Emacs 27 reimplements the API as generic functions and adds support for plists. It also deprecates `map-put` and the `testfn` argument of `map-elt` and `map-contains-key`.

These features were recently released as `map.el` version 2.0, and can be installed even on Emacs 25 and 26 via GNU ELPA: http://elpa.gnu.org/packages/map.html

yyoncho:
Updated as per latest review.

Fuco1:
Hey @yyoncho, sorry for the lack of feedback, I had vacation the last 10 days. I will process the changes ASAP, I really want this to land.

yyoncho:
@Fuco1 no problem, we had plenty of work to do on `lsp-mode` side.

yyoncho:
@Fuco1 did you had a chance to re-review the PR?

Fuco1:
Thanks to everyone involved.  I'm going to update the documentation now.

yyoncho:
Thank you! 

One question: previously we(lsp-mode team) were using some of the dash features that were not released(smart destructuring) and it was pretty common to get an issue report due to the fact that lsp-mode was installed afterwards and without updating dash. Even worse once bytecompiled against the old dash version the user had to do byte recompile in order to fix the issue. I want to avoid that situation so do you plan to do new release or eventually bump the package version (which I think might fix the problem I just described) or roll out a new release.

Fuco1:
@yyoncho Yes I'd like to release a new version with this feature.  Not sure if it helps as many people run from MELPA and there the version bump is "automatic".

Still we should be responsible and do a release when we get a major feature in.

I'm flying to Bulgaria on Sunday so hopefully I can get something done on the airport/airplane, but before then I need to finish all my "regular" work.

yyoncho:
@Fuco1 no pressure. I just wanted to have a clue whether it is planned since current release took quite some time. 

> Not sure if it helps as many people run from MELPA and there the version bump is "automatic".

We run from melpa(unless I am missing smt) but IMO the following scenario is not covered (please correct me if I am wrong).

1. User has old dash(I guess the majority of the emacs melpa users do have it). 
2. User wants to try lsp-mode and does M-x package-install lsp-mode
3. lsp-mode is downloaded but dash version is still the old one since we depend on 2.14.1 and current installed dash version is 2.14.1 as well so no need to update.
4. lsp-mode fails 

This is what I want to avoid since we had more than 30 bug reports when we were using smart destructuring when it was unreleased.

Enjoy your stay here in Bulgaria!

Fuco1:
Step 3 is incorrect.  If you install from MELPA your dash version will be, at minimum `20100101....` (I don't know when MELPA started exactly).

Still, even `20100101` is always going to be more than any dash version we ever release.  This will make the resolver satisfied and it will in fact never update any package.  Basically MELPA makes transitive upgrades impossible, you always have to manually select all your packages and all of their dependencies for upgrade.

I have had the same issues with smartparens or dired-hacks or some of my other smaller packages.  People in flycheck also had these issues... it's unfortunate, but I don't think preventable, unless you start to depend on the MELPA versions directly.

This is one of the reasons why I'm abandoning the whole package.el ecosystem and pushing towards https://github.com/raxod502/straight.el

Thanks! We're going to Plovdiv to check some "ruins" and then the mountains around Sofia :)

Issue #228, opened 2017-08-10
-<< / swiss-arrows?


alphapapa:
```el
(defmacro -<< (expr &rest forms)
  (let* ((result (gensym))
         (funcalls (cl-loop for form in forms
                            collect `(funcall #',(car form) ,@(cdr form) ,result))))
    
    `(when-let ((,result ,expr))
       (list ,@funcalls))))
```
e.g.:
```el
(-<< 3
     (+ 1)
     (format "Number: %s")
     (* 10))  ; => (4 "Number: 3" 30)

;; or

(-<< (listify-key-sequence (kbd "C-c m"))
     (event-modifiers)
     (event-basic-type))  ; => ((control) 99)
```

Fuco1:
We already have this except for functions:

    (funcall (-juxt '1+ (lambda (x) (format "Number %s" x)) (lambda (x) (* 10 x))) 3)

is the first example's equivalent.  I suppose we could add an anaphoric version /shrug.

Where is your notation coming from?

alphapapa:
I stumbled upon this recently:

https://github.com/rplevy/swiss-arrows

Then I had occasion to use a macro like this, so I thought I'd send it in and see what you thought.

alphapapa:
`-tee` might also be an appropriate name, since it's like the `tee` command.  Then `--tee` could be an anaphoric version.

alphapapa:
Here are some implementations that I've found useful:
```el
(defmacro -< (expr &rest forms)
  "Return a list of EXPR fed through FORMS, inserting it as the last item in each form.
EXPR is only evaluated once."
  (declare (indent defun))
  (let ((var (gensym)))
    `(let ((,var ,expr))
       (list ,@(--map (pcase it
                        ((pred symbolp) (list it var))
                        ((pred listp) (-snoc it var)))
                      forms)))))
(defmacro -<< (expr &rest forms)
  "Return a list of EXPR fed through FORMS, inserting it as the first item in each form.
EXPR is only evaluated once."
  (declare (indent defun))
  (let ((var (gensym)))
    `(let ((,var ,expr))
       (list ,@(--map (pcase it
                        ((pred symbolp) (list it var))
                        (`(,first . ,rest) `(,first ,var ,@rest)))
                      forms)))))

(defmacro --< (expr &rest forms)
  "Return a list of EXPR fed through FORMS.
EXPR is bound to `it' in each form and is only evaluated once."
  (declare (indent defun))
  `(let ((it ,expr))
[...]

PR #279, opened 2018-12-31
Add the branching arrows -<, -<<, --<, and -as-<.

malaggan:
Add parallel-threading macros inspired by https://github.com/rplevy/swiss-arrows. The general idea is the to complement the existing threading macros, which apply forms in sequence, to provide similar macros that apply forms in parallel. 

For example: `(-< 3 list 1+ 1- (* 2))` will produce `(4 2 6)`, and `(-< 3 + 1+ 1- (* 2))` will produce `12`. The syntax is similar to the sequential threading macros (->, ->>, etc.) except that an extra parameter is needed: the "collector" that will take the forms as arguments. This choice was made, instead of returning a list, to allow for short-circuiting special forms like `and` to be used.

The use case I've personally encountered that motivated me to implement this was the need to pass the same value to several predicates and combine these predicates using `and`:

```el
(and (listp element)
        (-> element length (>= 2))
        (-> element car atom)
        (-> element cadr alistp)
       element)
```

could be written as:

```el
(-< element and
      listp
      (-> length (>= 2))
      (-> car atom)
      (-> cadr alistp)
      identity))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ Functions reducing lists into single value.
 * [-inits](#-inits-list) `(list)`
 * [-tails](#-tails-list) `(list)`
 * [-common-prefix](#-common-prefix-rest-lists) `(&rest lists)`
+* [-common-suffix](#-common-suffix-rest-lists) `(&rest lists)`
 * [-min](#-min-list) `(list)`
 * [-min-by](#-min-by-comparator-list) `(comparator list)`
 * [-max](#-max-list) `(list)`
@@ -275,6 +276,10 @@ Functions pretending lists are trees.
 * [-some->](#-some--x-optional-form-rest-more) `(x &optional form &rest more)`
 * [-some->>](#-some--x-optional-form-rest-more) `(x &optional form &rest more)`
 * [-some-->](#-some---x-optional-form-rest-more) `(x &optional form &rest more)`
+* [-<](#--x-collector-rest-forms) `(x collector &rest forms)`
+* [-<<](#--x-collector-rest-forms) `(x collector &rest forms)`
+* [--<](#---x-collector-rest-forms) `(x collector &rest forms)`
+* [-as-<](#-as--x-collector-variable-rest-forms) `(x collector variable &rest forms)`
 
 ### Binding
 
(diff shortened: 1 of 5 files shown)


alphapapa:
https://github.com/magnars/dash.el/issues/228

malaggan:
> #228

Thank you for pointing me to this. I was not aware an issue was already open. How do you propose we proceed now?

alphapapa:
@malaggan D:  Did you change your mind?  I was looking forward to being able to use these.

README.md
```diff
@@ -1405,19 +1420,19 @@ other value (the body).
 Partition directly after each time `pred` is true on an element of `list`.
 
 ```el
-(-partition-after-pred #'oddp '()) ;; => '()
-(-partition-after-pred #'oddp '(1)) ;; => '((1))
-(-partition-after-pred #'oddp '(0 1)) ;; => '((0 1))
+(-partition-after-pred (function oddp) '()) ;; => '()
```
Fuco1: Can we get rid of these changes?

README.md
```diff
@@ -1405,19 +1420,19 @@ other value (the body).
...
 
 #### -partition-before-pred `(pred list)`
 
 Partition directly before each time `pred` is true on an element of `list`.
 
 ```el
-(-partition-before-pred #'oddp '()) ;; => '()
-(-partition-before-pred #'oddp '(1)) ;; => '((1))
-(-partition-before-pred #'oddp '(0 1)) ;; => '((0) (1))
+(-partition-before-pred (function oddp) '()) ;; => '()
```
Fuco1: Can we get rid of these changes?

README.md
```diff
@@ -2658,7 +2742,7 @@ expects a list with n items as arguments
 
 ```el
 (-map (-applify '+) '((1 1 1) (1 2 3) (5 5 5))) ;; => '(3 6 15)
-(-map (-applify (lambda (a b c) `(,a (,b (,c))))) '((1 1 1) (1 2 3) (5 5 5))) ;; => '((1 (1 (1))) (1 (2 (3))) (5 (5 (5))))
+(-map (-applify (lambda (a b c) (\` ((\, a) ((\, b) ((\, c))))))) '((1 1 1) (1 2 3) (5 5 5))) ;; => '((1 (1 (1))) (1 (2 (3))) (5 (5 (5))))
```
Fuco1: Can we get rid of these changes?

dash.el
```diff
@@ -1554,6 +1554,69 @@ and when that result is non-nil, through the next form, etc."
                    (--> ,result ,form))
                  ,@more))))
 
+(defmacro ---thread-each-and-collect (op x collector &optional forms)
```
Fuco1: use `dash--` prefix for private functions

dash.el
```diff
@@ -1554,6 +1554,69 @@ and when that result is non-nil, through the next form, etc."
...
+  "Private: Used by -< and -<<.
+
+Returned expr: (COLLECTOR (OP X form1) (OP X form2) ...).
+If no forms were passed then returned expr would be: (COLLECTOR X).
+X will be evaluated only once.
+"
+  (let ((value (make-symbol "value")))
+    `(let ((,value ,x))
+       (,collector ,@(--map (list op value it)
```
Fuco1: shouldn't `value` be unquoted?

malaggan: The whole sexp is unquoted: `,@(--map (list op value...`.

README.md
Fuco1: Yea, but I don't know why.  I think your emacs version and the previous person's who run that are different and that makes the inconsistencies.

We really should build the docs on the CI env.

dash.el
Fuco1: of course :) :+1:

PR #280, opened 2019-01-05
Add the branching arrows -<, -<<, --<, and -as-<. (resubmitted)

malaggan:
Note: I had to resubmit a new pull request instead of https://github.com/magnars/dash.el/pull/279 because the previous pull request cannot be reopened.

Add parallel-threading macros inspired by https://github.com/rplevy/swiss-arrows. The general idea is the to complement the existing threading macros, which apply forms in sequence, to provide similar macros that apply forms in parallel. 

For example: `(-< 3 list 1+ 1- (* 2))` will produce `(4 2 6)`, and `(-< 3 + 1+ 1- (* 2))` will produce `12`. The syntax is similar to the sequential threading macros (->, ->>, etc.) except that an extra parameter is needed: the "collector" that will take the forms as arguments. This choice was made, instead of returning a list, to allow for short-circuiting special forms like `and` to be used.

The use case I've personally encountered that motivated me to implement this was the need to pass the same value to several predicates and combine these predicates using `and`:

```el
(and (listp element)
        (-> element length (>= 2))
        (-> element car atom)
        (-> element cadr alistp)
       element)
```

could be written as:

```el
(-< element and
      listp
      (-> length (>= 2))
      (-> car atom)
      (-> cadr alistp)
      identity))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ Functions reducing lists into single value.
 * [-inits](#-inits-list) `(list)`
 * [-tails](#-tails-list) `(list)`
 * [-common-prefix](#-common-prefix-rest-lists) `(&rest lists)`
+* [-common-suffix](#-common-suffix-rest-lists) `(&rest lists)`
 * [-min](#-min-list) `(list)`
 * [-min-by](#-min-by-comparator-list) `(comparator list)`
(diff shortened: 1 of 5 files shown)


alphapapa:
See #228.

PR #282, opened 2019-01-28
Implement --doto

yyoncho:
Please let me know if you want me regenerate the docs. When I ran
./create-docs.sh it generates a lot of changes not related to this PR.


Fuco1:
This is fine, thanks!  I will run the doc gen and also probably set up a docker or something for that so we all use the same version of the utilities.

Issue #283, opened 2019-02-01
[Feature Request]Add new function `-shuffle`


cireu:
Since we have `-sort` to sort a list with fn. There should be a `-shuffle` to disorder the list

I try to wrote a function like 

```
(defun -shuffle (list)
  (declare (pure t) (side-effect-free t))
  (let* ((source (copy-sequence list))
         (random-nums (->> list
                           length
                           (number-sequence 1)
                           (-map #'random)
                           nreverse))
         result)
    (--each random-nums
      (setq source (-rotate it source))
      (push (car source) result)
      (!cdr source))
    result))
```

It generate a number-sequence first then use `random` to disorder and `nreverse` to reverse it.
Then use these random number to `-rotate` the list. collect the head and reset the tail to source
for next rotation.

Do you have faster implementation?

Fuco1:
Generally we don't care much for performance until proven necessary.  Elisp isn't a mathematical package so we're not epxecting you to work with lists of 10^5 orders.  For anything less any implementation is good enough.

Having said that, shuffling an array is actually a subtly tricky stuff and many people get it wrong.  Your algorithm also isn't exactly correct, although for practical purposes it might be enough.

You can read this blog https://www.i-programmer.info/programming/theory/2744-how-not-to-shuffle-the-kunth-fisher-yates-algorithm.html or this wikipedia page https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle for more details.

tl;dr we want to make sure any shuffle has equal probability of being generated.

Fuco1:
Oh wait nevermind, it's actually correct.

I thought `number-sequence` just generated `(10 10 10 10 10 10 ...)` which would then make the bias towards front element (since e.g rotation of 1 and 9 on an 8-length list is the same).  But it actually generates an ascending sequence.

In this case your algorithm is basically what I linked.  Sorry, I just woke up, coffee is still processing :D

cireu:
```
(defun -shuffle1 (list)
  (let ((source (copy-sequence list))
        (random-nums (->> list
                          length
                          (number-sequence 1)
                          (-map #'random)
                          nreverse))
        result)
    (--each random-nums
      (let* ((part1 (-take it source))
             (part2 (-drop it source))
             (target (pop part2)))
        (push target result)
        (setq source (nconc part1 part2))))
    result))
```
A rigorous impletation of Fish-Yate's shuffle alorithm

[...]

PR #286, opened 2019-02-01
Add new function `-shuffle`

cireu:
See https://github.com/magnars/dash.el/issues/283

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -2415,8 +2415,8 @@ See https://en.wikipedia.org/wiki/Fisher-Yates_shuffle for more details."
                           nreverse))
         result)
     (--each random-nums
-      (let* ((front (-take it list))
-             (rest (-drop it list))
+      (let* ((front (-take it source))
+             (rest (-drop it source))
              (target (pop rest)))
         (push target result)
         (setq source (nconc front rest))))
--- a/dash.el
+++ b/dash.el
@@ -2406,7 +2406,7 @@ if the first element should sort before the second."
 
 (defun -shuffle (list)
   "Shuffle the LIST, return a disordered list.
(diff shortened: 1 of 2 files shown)


dash.el
```diff
@@ -2404,6 +2404,22 @@ if the first element should sort before the second."
...
+Shuffling the list by using Fisher-Yates shuffle algorithm.
+See https://en.wikipedia.org/wiki/Fisher-Yates_shuffle for more details."
+  (let ((source (copy-sequence list))
+        (random-nums (->> (length list)
+                          (number-sequence 1)
+                          (-map #'random)
+                          nreverse))
+        result)
+    (--each random-nums
+      (-let [(front (target . rest)) (-split-at it source)]
```
Fuco1: Nice, even better than what I had :)

cireu:
Why CI test failed? Flycheck also complains that `dash--match`'s definition is void. But `dash--match` was defined at line 1849 @Fuco1

Fuco1:
You can execute `(random "dash")` to set the seed for the random generator, then it should always return the same sequence

```elisp
(progn
  (random "dash")
  (message "first %s" (-map (lambda (x) (mod x 100)) (-iterate #'random 10 10)))
  (random "dash") ;; comment out to see a different sequence
  (message "second %s" (-map (lambda (x) (mod x 100)) (-iterate #'random 10 10))))
```

cireu:
I  know, but it seems that `dash--match` is at line 1849, `-let` is at line 1916. And `-shuffle` is at line 2407

Fuco1:
@cireu I'm not sure then.  Probably you will have to expand the code or rewrite it without the `-let` macro :(

I've run into this problem before and I couldn't solve it but I should give it more time because it's quite annoying.

PR #290, opened 2019-03-10
Generalize `-rotate` for `|n|` greater than `(length list)`.

### LINKED ISSUE #284: (-rotate 7 '(1 2 3)) should work by moding 7 by length of list

leungbk:
Closes #284.

This works but is probably not idiomatic; I'd appreciate suggestions on how to tidy it up.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -921,9 +921,11 @@ See also: `-drop'"
   "Rotate LIST N places to the right.  With N negative, rotate to the left.
 The time complexity is O(n)."
   (declare (pure t) (side-effect-free t))
-  (if (> n 0)
-      (append (last list n) (butlast list n))
-    (append (-drop (- n) list) (-take (- n) list))))
+  (when list
+    (let* ((len (length list))
+           (n-mod-len (mod n len))
+           (new-tail-len (- len n-mod-len)))
+      (append (-drop new-tail-len list) (-take new-tail-len list)))))
 
 (defun -insert-at (n x list)
   "Return a list with X inserted into LIST at position N.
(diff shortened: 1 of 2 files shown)


dash.el
```diff
@@ -921,9 +921,11 @@ See also: `-drop'"
   "Rotate LIST N places to the right.  With N negative, rotate to the left.
 The time complexity is O(n)."
   (declare (pure t) (side-effect-free t))
-  (if (> n 0)
-      (append (last list n) (butlast list n))
-    (append (-drop (- n) list) (-take (- n) list))))
+  (and list
```
Fuco1: It's better to use `when` if there is only one condition followed by a body.

dash.el
```diff
@@ -921,9 +921,11 @@ See also: `-drop'"
...
+       (let* ((list-len (length list))
```
Fuco1: This could probably be called just `len`, it's understandable from context

dash.el
```diff
@@ -921,9 +921,11 @@ See also: `-drop'"
...
+              (n-mod-len (mod n list-len))
```
Fuco1: It's nice here that it gets rid of the special case for negatives as it automatically produces the corresponding positive offset <3

Fuco1:
@leungbk Not bad at all, just minor style issues and I'm happy to merge!

leungbk:
@Fuco1 Thanks for your suggestions, I've updated the PR accordingly.

Fuco1:
@leungbk 	Thanks! :rocket:

Issue #294, opened 2019-04-12
`hash?` expander evaluates its args twice.


cireu:
```
ELISP> (let* ((ht (make-hash-table :test #'equal))
              (fn (lambda (ht) (cl-incf (gethash 'a ht)) ht)))
         (puthash 'a 3 ht)
         (macroexpand-all '(-let (((&hash? 'a) (funcall fn ht)))
                             a)))
(let
    ((input0
      (funcall fn ht)))
  (let*
      ((a
        (if input0
            (progn
              (gethash 'a input0)))))
    a))


ELISP> (let* ((ht (make-hash-table :test #'equal))
              (fn (lambda (ht) (cl-incf (gethash 'a ht)) ht)))
         (puthash 'a 3 ht)
         (macroexpand-all '(-let [(&hash? 'a) (funcall fn ht)]
                             a)))
(let*
    ((a
      (if
          (funcall fn ht)
          (progn
            (gethash 'a
                     (funcall fn ht))))))
  a)
```

This may have problem with new `-let` expander `hash?`. For example, The both form will return 5

cireu:
It seems `dash-expand:hash?` also have problems. It evaluates its argument twice without protection, if you use `(-let (((&hash? 'a) (fn-generates-a-hash-table))) a)` or `(-let [(&hash? 'a) (fn-generates-a-hash-table)] a)`. And that function contains side-effect. It may cause problem.

I would like to know what solution you want, either modify `dash-expand:hash?`, wrap its body with `macroexp-let2` or change the behavior when `-let` meets pattern with only one placeholder. 

@yyoncho  @Fuco1

yyoncho:
IMO it should be unified with `-let(` but it is up to @Fuco1 to decide whether the potentially redundant `let` binding is a concern. I am not understanding `macroexp-let2` solution.

Fuco1:
The current behaviour is correct, we should make sure the expanders don't evaluate twice.

`-let` will use an intermediate variable if necessary.

```elisp
(-let [(&hash? 'a 'b) (funcall fn ht)]
  a)


(let*
    ((--dash-source-210--
      (funcall fn ht))
     (a
      (when --dash-source-210--
        (gethash 'a --dash-source-210--)))
     (b
      (when --dash-source-210--
        (gethash 'b --dash-source-210--))))
  a)
```

[...]

PR #296, opened 2019-04-13
Ensure `hash?` expander evaluates its arg only once.

cireu:
Close https://github.com/magnars/dash.el/issues/294


Fuco1:
Thanks!

PR #302, opened 2019-05-13
Remove mention of Marmalade

bbatsov:
It has been dead for a while now.


magnars:
Thanks! 👍🙂

basil-conto:
`README.md` is auto-generated, so this change should have been applied to `readme-template.md` and `dash-template.texi` instead. I have now done this in #309 along with some other fixes.

Issue #306, opened 2019-08-08
adding orElse for partial functions


ag91:
Hi there, thanks for the library: it has a lot of goodies!
I was working on some functionality I want to add to Emacs and I found out I needed an orElse (https://stackoverflow.com/questions/38863103/what-is-correct-way-to-use-operator-orelse-in-scala):
It basically recover on elisp functions that return `nil` for some inputs by running another elisp function on the same input. This is what I came with:

``` elisp
   (defun a (i) (if (< i 0) "first" nil))
   (defun b (i) (if (> i 0) "second" nil))
   (defun c (i) "third")
   (defun or-else (f g) `(lambda (i) (let ((x (,f i))) (or x (,g i)))))
   (funcall (or-else 'a 'b) -10)
   (funcall (or-else 'a (or-else 'b 'c)) 0)
```
Can this be an addition to your collection of useful functions? Or maybe this is already there and I did not see it?

Thanks again!

Fuco1:
The function seems useful and can be implemented directly as closure without using the backtick macro.  We could also make it take arbitrary number of functions not just two.

basil-conto:
This sounds like a more useful extension of `-orfn`:

```el
(defun -orfn (&rest preds)
  "Return a function which acts like `or' on the results of PREDS.
The returned function takes a variable number of ARGS.  When
called, it applies each n-ary function in PREDS to ARGS in turn
until one of them returns non-nil, then returns that value.  The
remaining PREDS are not called at all.  If all PREDS return nil,
the returned function also returns nil."
  (lambda (&rest args)
    (--some (apply it args) preds)))
```

If this is a welcome extension, I will submit a PR which makes other unary `dash-functional` definitions n-ary.

Fuco1:
I would prefer to keep the original `-orfn` definition since some people might rely on it returning `t` or `nil`.  We can call this `-somefn` perhaps?

magnars:
Agreed. Let’s not break people’s code. A new name is the way to go.

lør. 17. aug. 2019 kl. 22:53 skrev Matus Goljer <notifications@github.com>:

basil-conto:
Of course I will ultimately heed the maintainers' wishes, but here are some arguments for extending `-orfn` instead of introducing yet another variation of an existing model, which I am personally against:

[...]

PR #308, opened 2019-08-29
Prefer n-ary over unary combinators where possible

basil-conto:
This is a more focused and efficient rewrite of PR #72 in order to address issue #306.

Cc: @vapniks

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,8 +373,9 @@ Functions that manipulate and compose other functions.
 * [`-juxt`](#-juxt-rest-fns) `(&rest fns)`
 * [`-compose`](#-compose-rest-fns) `(&rest fns)`
 * [`-applify`](#-applify-fn) `(fn)`
-* [`-on`](#-on-operator-transformer) `(operator transformer)`
-* [`-flip`](#-flip-func) `(func)`
+* [`-on`](#-on-op-trans) `(op trans)`
+* [`-flip`](#-flip-fn) `(fn)`
+* [`-rotate-args`](#-rotate-args-n-fn) `(n fn)`
 * [`-const`](#-const-c) `(c)`
 * [`-cut`](#-cut-rest-params) `(&rest params)`
 * [`-not`](#-not-pred) `(pred)`
@@ -1193,7 +1194,7 @@ Return the smallest value from `list` of numbers or markers.
 Take a comparison function `comparator` and a `list` and return
 the least element of the list by the comparison function.
 
-See also combinator [`-on`](#-on-operator-transformer) which can transform the values before
+See also combinator [`-on`](#-on-op-trans) which can transform the values before
 comparing them.
 
 ```el
@@ -1217,7 +1218,7 @@ Return the largest value from `list` of numbers or markers.
 Take a comparison function `comparator` and a `list` and return
 the greatest element of the list by the comparison function.
 
-See also combinator [`-on`](#-on-operator-transformer) which can transform the values before
+See also combinator [`-on`](#-on-op-trans) which can transform the values before
 comparing them.
 
 ```el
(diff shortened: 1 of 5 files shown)


basil-conto:
I have marked this PR as a draft because I have not yet got around to auditing relevant docs and tests.

I have also benchmarked `-on` but not the other combinators, and am not yet convinced trading generality for performance is worth it.

basil-conto:
> I have also benchmarked `-on` but not the other combinators, and am not yet convinced trading generality for performance, as was suggested in #72, is worth it.

Indeed I am now of the opinion that generality in these combinators is a more desirable trait than maintaining identical performance with their previous versions; see below.

Here are some benchmarks inspired by #72 that use `-on` as an example:

```el
;;; on.el --- benchmarks for -on -*- lexical-binding: t -*-

(defun on-binary (op trans)
  (lambda (x y)
    (funcall op (funcall trans x) (funcall trans y))))

(defun on-biniadic (op trans)
  (lambda (x y &rest args)
    (if args
        (apply op (funcall trans x) (funcall trans y)
               (mapcar trans args))
      (funcall op (funcall trans x) (funcall trans y)))))

(defun on-binariadic (op trans)
  (lambda (x y &rest args)
    (apply op (mapcar trans `(,x ,y ,@args)))))

(defun on-variadic (op trans)
  (lambda (&rest args)
    (apply op (mapcar trans args))))

(dolist (on '(on-binary on-binary on-biniadic on-binariadic on-variadic))
  (let ((list (mapcar #'number-to-string (number-sequence 0 10000)))
        (cmp  (funcall on #'> #'string-to-number)))
    (garbage-collect)
    (apply #'message "%-14s %9.6f %2d %.3f" on
           (benchmark-run 100
             (sort list cmp)))))

;;; on.el ends here
```

Running them in the following Emacs version:

```
In GNU Emacs 27.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, Xaw3d scroll bars)
 of 2019-08-27 built on thunk
Repository revision: 4118297ae2fab4886b20d193ba511a229637aea3
Repository branch: master
Windowing system distributor 'The X.Org Foundation', version 11.0.12004000
System Description: Debian GNU/Linux bullseye/sid

Configured using:
 'configure 'CC=ccache gcc' 'CFLAGS=-O2 -march=native' --config-cache
 --prefix=/home/blc/.local --with-mailutils --with-x-toolkit=lucid
 --with-modules --with-file-notification=yes --with-x'

Configured features:
XAW3D XPM JPEG TIFF GIF PNG RSVG SOUND GPM DBUS GSETTINGS GLIB NOTIFY
INOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF
XFT ZLIB TOOLKIT_SCROLL_BARS LUCID X11 XDBE XIM MODULES THREADS
LIBSYSTEMD JSON PDUMPER LCMS2 GMP
```

Gives the following results:

```
emacs -batch -f batch-byte-compile on.el && emacs -script on.elc
on-binary       0.027191  0 0.000
on-binary       0.015568  0 0.000
on-biniadic     0.015991  0 0.000
on-binariadic   0.065449  5 0.040
on-variadic     0.063849  5 0.040
```

The extra consing involved in the variadic implementations clearly slows down the returned function.

One solution would be to go with the implementation of `on-biniadic`, in which the `&rest` arglist is only allocated if the returned function is called with more than two arguments. This implementation exhibits the same performance as the original `on-binary` implementation. This is a win-win, because it improves on the current implementation without sacrificing performance.

My argument, however, is that we ought to go with `on-variadic`, despite the performance hit. Here is an outline of my reasoning:

* The performance of `&rest` args is inherent to Elisp and neither specific to nor the responsibility of the Dash library. Until Emacs core is further optimised, everyone must eventually pay this price.
* `dash-functional` is a library providing higher order combinators for convenience.
  * Such functions are inherently prone to lower performance, especially in a language like Elisp.
  * Performance-critical Elisp shouldn't be using these combinators to begin with.
  * The purpose of such functions is to strive for greater expressiveness, composability, conciceness, elegance, etc.
    * Restricting the calling conventions of such functions goes against all of these goals.
* The benchmark above shows very reasonable performance for repeatedly sorting a large list, even if slower than the original implementation. I think we can live with this.

WDYT?

basil-conto:
I intend to mark this PR as ready to review after the issues that PR #309 addresses are fixed, so that this branch can be rebased.

vapniks:
I have spent the past few weeks learning txr lisp which has it's own (n-ary) versions of these combinators which I found useful. I wrote some txr code for comparing function runtimes and simple code complexity measures (maxdepth, number of symbols, etc), and was thinking that these measures might also be useful for [elsa](https://github.com/emacs-elsa/Elsa). 

In principle there ought to be some empirical way of determining the best trade off between runtime & complexity under different situations, e.g. by doing some data analysis on large amounts of lisp code pulled from github, but as it stands I agree with the prognosis of @basil-conto

basil-conto:
> I have marked this PR as a draft because I have not yet got around to auditing relevant docs and tests.

> I intend to mark this PR as ready to review after the issues that PR #309 addresses are fixed, so that this branch can be rebased.

I have now adapted the docs and tests, and have rebased the branch on PR #309 so that the tests can pass on all Emacs versions. Ideally if that PR is merged I will then mark this PR as ready to review.

basil-conto:
> I intend to mark this PR as ready to review after the issues that PR #309 addresses are fixed, so that this branch can be rebased.

I have now rebased this branch on top of the merged #309 and have marked this PR as ready for review.

basil-conto:
> My argument, however, is that we ought to go with `on-variadic`, despite the performance hit.

Of course, it's possible to optimise `on-variadic` for the two argument case, to avoid the `mapcar`+`apply` overhead, if desired. ;)

I think I'll wait until after #356 is addressed to implement that, though.

README.md
```diff
@@ -2738,40 +2738,44 @@ See `srfi-26` for detailed description.
...
+Return a function that is the logical negation of predicate `pred`.
+When called, the returned variadic function passes its arguments
+to `pred` and returns t if `pred` returns nil, otherwise nil.
 
 ```el
-(funcall (-not 'even?) 5) ;; => t
-(-filter (-not (-partial '< 4)) '(1 2 3 4 5 6 7 8)) ;; => '(1 2 3 4)
+(funcall (-not #'even?) 5) ;; => t
+(-filter (-not (-partial #'< 4)) '(1 2 3 4 5 6 7 8)) ;; => '(1 2 3 4)
+(funcall (-not #'>) 1 2) ;; => t
```
Fuco1, line 2748: I just tried something like this yesterday and was sad it doesn't work with a binary predicate.  Definitely a welcome change.

Fuco1:
I agree completely that the performance here is secondary.  First come the users and then we can optimize what is reported to be slow.  This has so far always turned out to be the correct way forward.

PR #309, opened 2019-08-30
Fix doc and test harnesses

basil-conto:
TL;DR:
* Byte-compiled `-third-item` is broken in several Emacs versions.
* Test harness currently fails in Emacs 27 due to unconditional loading of officially deprecated `cl.el`.
* Doc generation currently incorrectly replaces `'()` with `'nil`.
* Mentions of Marmalade were not properly removed in #302.

Please see commit messages for more details.


basil-conto:
I will continue trying to fix the following error later:

`Error: the function `gv--defsetter' might not be defined at runtime.`

basil-conto:
[BTW, my favourite solution would be to drop support for Emacs 24.1 and 24.2, if possible ;) ]

basil-conto:
Phew, I think I got there in the end. In the process I discovered and fixed a bug in several Emacs versions:

1. `emacs -Q -batch -f batch-byte-compile dash.el`
2. `emacs -Q -l dash.elc`
3. ```el
   (-third-item ())
   ```
   <kbd>C-j</kbd>

This results in a `(void-function -third-item)` error because `dash.el` currently incorrectly assumes that loading `cl.el` at byte-compile-time will define `caddr` at runtime.

A future PR will improve usage of built-in and `cl-lib` `c[ad]+r` accessors, but this PR is just about getting byte-compilation, testing, and doc generation working consistently across the board.

dev/examples-to-docs.el
```diff
@@ -153,8 +153,8 @@ FUNCTION may reference an elisp function, alias, macro or a subr."
 
 (defun simplify-quotes ()
   (goto-char (point-min))
-  (while (search-forward "(quote nil)" nil t)
-    (replace-match "'()"))
+  (while (re-search-forward (rx (| "'nil" "(quote nil)")) nil t)
```
Fuco1: Tiny nitpick: I'd prefer `or` over `|`.  Also in the other files.

```suggestion
  (while (re-search-forward (rx (or "'nil" "(quote nil)")) nil t)
```

basil-conto: Done.

Fuco1:
Super!

PR #313, opened 2019-09-18
Make -inits not destroy its argument

SwiftLawnGnome:
Fixes #312

PR #264, opened 2018-01-21
Improve folds

basil-conto:
Please review each commit individually. The key points are:

* Make recursive folds iterative to avoid exceeding `max-lisp-eval-depth`.
* Handle an empty list argument properly in `-reductions` and `-reductions-r`.

The docs will probably need to be regenerated if any other commits are added in the meantime.

PR #314, opened 2019-09-19
Add anaphoric versions of -reductions et al.

basil-conto:
* `dash.el` (`--reductions-from`, `--reductions`)
(`--reductions-r-from`, `--reductions-r`): New anaphoric macros.
(`-reductions-from`, `-reductions`, `-reductions-r-from`, `-reductions-r`):
Use them, respectively.
(`-tails`, `-running-sum`, `-running-product`):
* `dev/examples.el` (`-reductions-from`, `-reductions-r-from`)
(`-reductions`, `-reductions-r`): Quote function symbols as such.

* `README.md`:
* `dash.info`:

Re: #313.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,10 +151,10 @@ Functions returning a modified copy of the input list.
 ### Reductions
 
 
-Functions reducing lists into single value.
+Functions reducing lists to a single value (which may also be a list).
 
-* [-reduce-from](#-reduce-from-fn-initial-value-list) `(fn initial-value list)`
-* [-reduce-r-from](#-reduce-r-from-fn-initial-value-list) `(fn initial-value list)`
+* [-reduce-from](#-reduce-from-fn-init-list) `(fn init list)`
+* [-reduce-r-from](#-reduce-r-from-fn-init-list) `(fn init list)`
 * [-reduce](#-reduce-fn-list) `(fn list)`
 * [-reduce-r](#-reduce-r-fn-list) `(fn list)`
 * [-reductions-from](#-reductions-from-fn-init-list) `(fn init list)`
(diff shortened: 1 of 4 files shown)


basil-conto:
Marking this PR as draft because it is based on the unmerged PR #309.

Fuco1:
This PR is pretty hard to review for me.  There are two things going on:

1. We implemented the anaphoric versions
2. The non-anaphoric implementations now use the anaphoric versions

Do you think it would be too much effort to split it into two patches?  I guess it would take a lot of time now since it seems pretty intertwined.

I'm just a bit worried that we might've forgotten something or some function is not covered by the tests.  Maybe adding tests for the anaphoric versions would be less work.

magnars:
Adding tests for the anaphoric versions is a must-do anyway. We need it
both for regression testing and for the examples. Always have at least one
anaphoric and one non-anaphoric example in the first three to ensure we
have nice docs.

tor. 19. sep. 2019 kl. 20:10 skrev Matus Goljer <notifications@github.com>:

basil-conto:
@Fuco1 

> Do you think it would be too much effort to split it into two patches? I guess it would take a lot of time now since it seems pretty intertwined.

No, splitting the patch would be trivial for me; I'll try to get around to it later today.

All the current patch does is move the logic from the body of the functions to the body of the new anaphoric macros. It is trivial to duplicate the logic in the macros in one patch, and remove the old duplicate logic from the body of the functions in a second patch. I thought a single "move" operation would be simpler than a "copy then delete", but if the latter aids review, then I'm happy to oblige.

> I'm just a bit worried that we might've forgotten something or some function is not covered by the tests.

Since the functions are implemented in terms of their anaphoric counterparts, the existing tests automatically cover both.

> Maybe adding tests for the anaphoric versions would be less work.

I'm happy to add tests explicitly calling the anaphoric macros, but not because it would be less work. ;)

@magnars 

> Adding tests for the anaphoric versions is a must-do anyway. We need it both for regression testing and for the examples. Always have at least one anaphoric and one non-anaphoric example in the first three to ensure we have nice docs.

As stated above, the anaphoric versions are already being regression tested. That said, I agree that it's better to target them explicitly, and examples are always nice. Will add more tests in the next patch iteration.

magnars:
> As stated above, the anaphoric versions are already being regression tested.

This is true only as long as the functions are implemented in terms of their anaphoric counterparts, which might not always be the case. :-) Thanks for adding the explicit tests! In case I wasn't clear, only the three first examples are used in the docs, so let's try to make sure one of each is used there.

basil-conto:
>> As stated above, the anaphoric versions are already being regression tested.

> This is true only as long as the functions are implemented in terms of their anaphoric counterparts, which might not always be the case. :-)

Yes, that's why I agreed it is better to target the macros explicitly.

> In case I wasn't clear, only the three first examples are used in the docs, so let's try to make sure one of each is used there.

Understood, thanks.

basil-conto:
Hm, in the process of tidying these definitions I noticed an inconsistency that my PR #264 covered up but didn't fix:

```el
(-reduce #'vector ())         ; => []
(-reduce #'vector '(1))       ; => 1
(-reduce #'vector '(1 2))     ; => [1 2]

(-reductions #'vector ())     ; => ()
(-reductions #'vector '(1))   ; => (1)
(-reductions #'vector '(1 2)) ; => (1 [1 2])
```

As you can see, `-reductions` is inconsistent with `-reduce` w.r.t. an empty input list. It was also inconsistent before #264, returning `(nil)` instead of `([])`.

I intend to fix this as well in this PR.

PR #317, opened 2019-09-27
Avoid `dash--get-expand-function` create non-existing symbol.

cireu:


Fuco1:
If this returns nil won't it break the code?  Or is that the point, to let the user know something is not right?

In that case we should probably also add a test for this behaviour so we keep it in the future.

cireu:
Still need some research on corner case, currently have no idea on how to test it.

basil-conto:
> If this returns nil won't it break the code?

No, because all current uses of `dash--get-expand-function` expect it to return a symbol satisfying `functionp`.

> Or is that the point, to let the user know something is not right?

This code does not seem to have been designed with that in mind.

> In that case we should probably also add a test for this behaviour so we keep it in the future.

This is part of the extensible destructuring logic added in #277, which is already being tested to an extent.  I think the general logic needs to be refactored a little before it can be more deeply and effectively tested.

cireu:
BTW, can we use `pcase` in dash.el to reduce boilderplates in `dash--match`?

Fuco1:
@cireu No pcase please.

@basil-conto Yea, I think I wasn't clear in my response.  I don't understand the purpose of this PR:

- what does it solve
- how does it break or not break existing code

basil-conto:
> what does it solve

It avoids interning arbitrary and unused symbols, i.e. it keeps the environment cleaner.

> how does it break or not break existing code

It should have no effect on existing code.

This PR gets a +1 from me, but I would prefer if the commit message were clearer @cireu, e.g. `Avoid interning unused symbols in destructuring`. More generally, widespread Git convention is to limit the first line of the message to 50 characters and leave any further details for the body of the message.

basil-conto:
Cc: @yyoncho

cireu:
@basil-conto  Check this, changes was mentioned in commit message body.

basil-conto:
@cireu 

> Check this, changes was mentioned in commit message body.

LGTM, thanks!

Fuco1:
> No, because all current uses of dash--get-expand-function expect it to return a symbol satisfying functionp.

I didn't get this at first, but now I understand what you ment.

LGTM as well, if nobody has anything to add I'll merge this later today/tonight.

cireu:
Are you ready for this PR? @Fuco1

Fuco1:
Thanks for reminding me @cireu and sorry for the wait!

Issue #315, opened 2019-09-20
Change the first line of the doc string to a complete sentence


mihaiolteanu:
The doc string in the echo area only shows the first line of the documentation string. But that line in the case of `dash` functions is not a complete and standalone sentence describing what the function does so it is close to useless. Or annoying, if you will.

Example for `-zip`, but it is the same for all the others (i.e. line ends in mid-sentence):
![screenshot1568982970](https://user-images.githubusercontent.com/8273519/65316816-4f69d980-dba3-11e9-8de2-f6d9c7a38476.png)

See also "Tips for Documentation Strings" in the Emacs Lisp Info pages
     
> The first line of the documentation string should consist of one or
> two complete sentences that stand on their own as a summary.  ‘M-x
> apropos’ displays just the first line, and if that line’s contents
> don’t stand on their own, the result looks bad.  In particular,
> start the first line with a capital letter and end it with a
> period.
> 
> For a function, the first line should briefly answer the question,
> “What does this function do?” For a variable, the first line should
> briefly answer the question, “What does this value mean?”
> 
> Don’t limit the documentation string to one line; use as many lines
> as you need to explain the details of how to use the function or
> variable.  Please use complete sentences for the rest of the text
> too.

basil-conto:
Indeed, this breaks from a good convention for no gain. I have been slowly chiselling away at Dash docstrings, e.g. in #308, but have not yet mustered the patience or courage to fix all of them in one go. I'll eventually get around to this, but no promises as to how soon... In the meantime, feel free to submit patch(es).

PR #318, opened 2019-10-10
Update/cleanup the docstrings (#315)

mihaiolteanu:
Attempted fixes:
- Insert double space between sentences
- Form a complete sentence as the first line in the docstring
- Mention all arguments and describe their purpose
- Rewording or rewriting the docstring from scratch

I'm assuming there will be different needs and/or opinions regarding the correct format of the docstrings, so this is mostly a proposal. It can be rejected, criticized, changed, etc.
I'm mostly interested to find out what preferences are there regarding the docstrings. Some of them can get quite wordy. It is a challenge to fill the gist of the function in one short and sweet sentence. 

I've kept the "Anaphoric form of...." references. I've moved the documentation of the captured variables inside the anaphoric macro docstring itself and removed it from the classic one. I'm not sure about the empty lines. Should an empty line exist between the first sentence that fills the first line and the second sentence or paragraph?

I'm not fully pleased with the wording for the `reduce` family of macros and functions.

Things can be improved a lot. It's a start, though. I wanted to start the discussion for this somewhere, somehow.

Mihai


ppareit:
This also reduces all those flycheck warnings. Very helpful. Hope this gets pulled.

dash.el
```diff
@@ -65,7 +65,9 @@ special values."
   `(setq ,list (cdr ,list)))
 
 (defmacro --each (list &rest body)
-  "Anaphoric form of `-each'."
+  "Eval BODY with symbol `it' bound to each element of the LIST.
+Return nil, used for side-effects only.
+Anaphoric form of `-each'."
```
Fuco1: I think a bit cleaner (from the perspective of reading the source code) way would be to put an empty line after the first line and before the "reference link" on the last.

``` elisp
"Eval BODY with symbol `it' bound to each element of the LIST.

Return nil, used for side-effects only.

Anaphoric form of `-each'."
```

This draws the attention to the first line better which is the quick summary.  On some functions we also say `See also: ....` which also is on its own line so it would be a bit more consistent.

WDYT?

dash.el
```diff
@@ -111,14 +118,18 @@ Note: `it' is not required in each form."
 (defun -each-indexed (list fn)
   "Call (FN index item) for each item in LIST.
 
-In the anaphoric form `--each-indexed', the index is exposed as symbol `it-index'.
+In the anaphoric form `--each-indexed', the index is exposed as
```
Fuco1: I think this sentence can be removed from here and move to the documentation of `--each-indexed` along with the description (the same way you did for the other anaphora).

Fuco1:
I personally prefer the empty line after the first summary line, but we should do what the rest of the file does to be consistent.  I suspect there isn't much of consistency though.

I very much like this initiative, often I'm annoyed having to click through from the anaphora to the other function to read the docstring.  My only concern is how do we make sure both are updated when the function/macro changes.

dash.el
mihaiolteanu: I agree that it's clearer with extra spaces. I like it better also. The question is if it doesn't add too much white-noise and if it's normal to have extra whitespace all over the place in the doc-strings.

Fuco1: I think it's pretty normal in Elisp world.  On some of my projects I have docstrings spaning multiple pages: https://github.com/Fuco1/smartparens/blob/9738360eb2afb58b4c21815f9d5c793b8125f540/smartparens.el#L2292

mihaiolteanu:
@Fuco1, I've addressed those issues from the review (add blank lines between paragraphs and remove the captured variables references in the classic functions).

 I've force-pushed some more docstring changes. I'll keep adding these until all are covered and/or we reach an agreement as to the best possible form. I'm not assuming this is the perfect wording or structure.

Thanks for your reply.

Fuco1:
A note on reviews, it's better not to force-push but just pile the commits one on top of another.  It helps to see what is the progression and where and how (and if even) things were fixed.  At the end either you or I can squash all the commits into a reasonable format.

As for the content, I'll review when I have a break.

mihaiolteanu:
I don't have time to waste if there is no interest in this.

Thanks for your time.

PR #321, opened 2019-10-30
Add missing indent declaration for ‘-some->’ and siblings

### LINKED ISSUE #319: missing indent declaration for ‘-some->’ and siblings
it seems the following macros could use a `(declare (indent 1))` since they all take one ‘special’ argument:

- `-some->`
- `-some->>`
- `-some-->`

wbolster:
The `->`, `->>`, and `-->` threading macros have short names, and they align
all arguments vertically, including the first one.

While the `-some->`, `-some->>`, and `-some-->` macros have similar
behaviour, they actually have quite long names, resulting in code that
looks like this:

```elisp
  (-some->> '(1 2 3)
            (do-something)
            (and-something-else))
```
Adding a `(declare (indent 1))` declaration for those variants
results in this indentation instead:

```elisp
  (-some->> '(1 2 3)
    (do-something)
    (and-something-else))
```

This is arguably better since the `-some-*` versions of these threading
macros actually treat their first argument a bit special: if the
expression evalutes to nil, the rest won't even run, and this
indentation makes the whole expression look more like a conditional
form.

Fixes #319.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1536,7 +1536,8 @@ VARIABLE to the result of the first form, and so forth."
 (defmacro -some-> (x &optional form &rest more)
   "When expr is non-nil, thread it through the first form (via `->'),
 and when that result is non-nil, through the next form, etc."
-  (declare (debug ->))
+  (declare (debug ->)
+           (indent 1))
   (if (null form) x
     (let ((result (make-symbol "result")))
       `(-some-> (-when-let (,result ,x)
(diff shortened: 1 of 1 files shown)


Fuco1:
That's a good point about the special treatment of the first arg.  Thanks!

wbolster:
tbh technically my reasoning is cheating. the semantics of the `-some` threading macros are that every arg is both a condition *and* normal code, not just the first. ;)

but even then i argue that the first arg is special because it is the starting point for the rest of the chain.

and the long name without the indentation change gives really weird deeply indented code compared to the both functional and visual elegance of the sibling `->` threading macros.

PR #323, opened 2019-11-08
Fix typos

tarsius:


Fuco1:
Apparently occurrence is a tricky word :D

PR #325, opened 2019-11-11
regenerate docs

wbolster:
i ran `pre-commit.sh` and kept only the relevant changes it made, since a different `emacs`/`makeinfo` version on my machine seems to produce slightly different results, giving weird diffs like this:

```diff
-(-partition-after-pred #'odd? '())
+(-partition-after-pred (function odd?) '())
```

(i skipped all of these weird changes)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1417,9 +1417,9 @@ other value (the body).
 Partition directly after each time `pred` is true on an element of `list`.
 
 ```el
-(-partition-after-pred #'odd? '()) ;; => '()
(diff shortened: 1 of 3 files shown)


wbolster:
outdated and no longer relevant, it seems. closing.

basil-conto:
Sorry this fell through the cracks, and thanks for paying attention to these things!

wbolster:
no worries. thanks for maintaining this library. :hugs:  it made elisp much more approachable for me; without it i would likely not be maintaining any serious melpa packages.

PR #336, opened 2020-05-18
Introduce -defun

yyoncho:
- I used `-lambda` as a base, tried several versions with extracting the common
code but neither of them made the code more readable(IMO).

- Removed the restrictions against doing `(-lambda () ...)` which does not seem
to needed.


ericdallo:
LGTM, maybe add some link/docs on the README on how to use?

dash.el
```diff
@@ -2166,7 +2164,38 @@ See `-let' for the description of destructuring mechanism."
...
+                           def-body)))
+  (cond
+   ;; no destructuring, so just return regular defun to make things faster
+   ((-all? 'symbolp match-form)
+    `(defun ,defun-name ,match-form ,@body))
+   (t
+    (let* ((inputs (--map-indexed (list it (make-symbol (format "input%d" it-index))) match-form)))
+      ;; TODO: because inputs to the defun are evaluated only once,
+      ;; -let* need not to create the extra bindings to ensure that.
+      ;; We should find a way to optimize that.  Not critical however.
```
basil-conto, line 2197: Can you please give an example of the optimisation that this comment is referring to?

dev/examples.el
```diff
@@ -1259,6 +1259,19 @@ new list."
...
+        (+ x y))
+
+      (-map #'dash-plus '((1 2) (3 4) (5 6)))) => '(3 7 11)
+
+    (progn
+      (-defun a+b ((&plist :a a :b b))
+        (+ a b))
+
+      (-map 'a+b '((:a 1 :b 2) (:a 3 :b 4) (:a 5 :b 6)))) => '(3 7 11))
+
```
basil-conto, line 1274: I don't know if it's worth shortening these examples to make them clearer, but keep in mind that the `create-docs.sh` script is going to flatten these examples into a single line.

basil-conto:
@ericdallo 

> maybe add some link/docs on the README on how to use?

For every `defexamples` entry in `dev/examples.el`, the `create-docs.sh` script adds the relevant function's docstring and its first few examples to `README.md`.

dash.el
```diff
@@ -2166,7 +2164,38 @@ See `-let' for the description of destructuring mechanism."
...
+has the usual semantics of `defun'.  Furthermore, these get
+translated into normal `defun', so there is no performance
+penalty.
+
+See `-let' for the description of destructuring mechanism."
+  (declare (doc-string 3) (indent defun)
+           (debug (&define sexp
+                           [&optional stringp]
+                           [&optional ("interactive" interactive)]
+                           def-body)))
```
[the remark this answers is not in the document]
yyoncho: thank you, good catch.

dash.el
```diff
@@ -2156,8 +2156,6 @@ See `-let' for the description of destructuring mechanism."
                            [&optional ("interactive" interactive)]
                            def-body)))
   (cond
-   ((not (consp match-form))
-    (signal 'wrong-type-argument "match-form must be a list"))
```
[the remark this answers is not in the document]
yyoncho: Make sense, I am adding this to both lambda and defun.

dash.el
yyoncho: I have copied the comment from above. I am not sure what it refers to.

dev/examples.el
yyoncho: I updated the example to call the `-defun` directly.

yyoncho:
@basil-conto addressed comments and force-pushed.

dash.el
```diff
@@ -2146,18 +2146,18 @@ such that:
   (-lambda (x y ...) body)
 
 has the usual semantics of `lambda'.  Furthermore, these get
-translated into normal lambda, so there is no performance
+translated a into normal `lambda', so there is no performance
```
basil-conto, line 2149: "...translated into a normal..."

dash.el
```diff
@@ -2146,18 +2146,18 @@ such that:
...
 penalty.
 
-See `-let' for the description of destructuring mechanism."
+See `-let' for a description of destructuring mechanism."
```
basil-conto, line 2152: Nit: "...of the destructuring mechanism."

dash.el
```diff
@@ -2166,7 +2166,41 @@ See `-let' for the description of destructuring mechanism."
       ;; TODO: because inputs to the lambda are evaluated only once,
       ;; -let* need not to create the extra bindings to ensure that.
       ;; We should find a way to optimize that.  Not critical however.
-      `(lambda ,(--map (cadr it) inputs)
+      `(lambda ,(-map #'cadr inputs)
+         (-let* ,inputs ,@body))))))
+
+(defmacro -defun (name match-form &rest body)
+  "Return a `defun' NAME which destructures its input as MATCH-FORM and executes BODY.
```
basil-conto, line 2173: "Define NAME as a function which..."

For consistency with the docstring of `defun`, and because a `defun` isn't something that can be "returned" in this context.

dash.el
```diff
@@ -2166,7 +2166,41 @@ See `-let' for the description of destructuring mechanism."
...
+
+Note that you have to enclose the MATCH-FORM in a pair of parens,
+such that:
+
+  (-defun (x) body)
+  (-defun (x y ...) body)
+
+has the usual semantics of `defun'.  Furthermore, these get
+translated a into normal `defun', so there is no performance
```
basil-conto, line 2182: "...translated into a normal..."

dash.el
```diff
@@ -2166,7 +2166,41 @@ See `-let' for the description of destructuring mechanism."
...
+penalty.
+
+See `-let' for a description of destructuring mechanism."
```
basil-conto, line 2185: Nit: "...of the destructuring mechanism."

dash.el
```diff
@@ -2166,7 +2166,41 @@ See `-let' for the description of destructuring mechanism."
...
+   ((nlistp match-form)
+    (signal 'wrong-type-argument (list #'listp match-form)))
+   ;; no destructuring, so just return regular defun to make things faster
+   ((-all? 'symbolp match-form)
+    `(defun ,name ,match-form ,@body))
+   (t
+    (let* ((inputs (--map-indexed (list it (make-symbol (format "input%d" it-index))) match-form)))
+      ;; TODO: because inputs to the defun are evaluated only once,
+      ;; -let* need not to create the extra bindings to ensure that.
+      ;; We should find a way to optimize that.  Not critical however.
```
basil-conto, line 2202: If this comment is not relevant to `-defun` then there is no point in cargo-culting it from `-lambda`.

dash.el
```diff
@@ -2166,7 +2166,41 @@ See `-let' for the description of destructuring mechanism."
...
+                           [&optional ("interactive" interactive)]
+                           def-body)))
+  (cond
+   ((nlistp match-form)
+    (signal 'wrong-type-argument (list #'listp match-form)))
+   ;; no destructuring, so just return regular defun to make things faster
+   ((-all? 'symbolp match-form)
+    `(defun ,name ,match-form ,@body))
+   (t
+    (let* ((inputs (--map-indexed (list it (make-symbol (format "input%d" it-index))) match-form)))
```
basil-conto, line 2199: No need for `let*` here; it can be a `let`.

dash.el
yyoncho: I can't tell if it is relevant or not, but if you or the maintainers want I could delete it.

yyoncho:
> > addressed comments and force-pushed.
> 
> Thanks, but some of the docs still contain mistakes; see below.

Sorry about that, please check the newly force pushed version(hopefully I have fixed everything). 

> By the way, `dash.el` is now included in GNU ELPA and is copyrighted by the Free Software Foundation. Do you have a copyright assignment on file?

I don't have an assignment.

basil-conto:
> I don't have an assignment.

Would you be willing to? You only need to do it once for all FSF-copyrighted Emacs packages, including Emacs core. It involves filling out one of the forms under http://git.savannah.gnu.org/cgit/gnulib.git/tree/doc/Copyright and emailing it to assign@gnu.org. They'll tell you what you need to do from there.

dash.el
```diff
@@ -2166,8 +2166,42 @@ See `-let' for the description of destructuring mechanism."
       ;; TODO: because inputs to the lambda are evaluated only once,
       ;; -let* need not to create the extra bindings to ensure that.
       ;; We should find a way to optimize that.  Not critical however.
-      `(lambda ,(--map (cadr it) inputs)
-         (-let* ,inputs ,@body))))))
+      `(lambda ,(-map #'cadr inputs)
+         (-let ,inputs ,@body))))))
```
basil-conto, line 2170: Changing this to `-let` would be a breaking change, just leave the `-let*` in place. In my previous review I was referring to the surrounding `let*` which can be safely turned into a `let`.

dash.el
```diff
@@ -2166,8 +2166,42 @@ See `-let' for the description of destructuring mechanism."
...
+
+(defmacro -defun (name match-form &rest body)
+  "Define NAME as a function which which destructures its input as MATCH-FORM and executes BODY.
```
basil-conto, line 2173: Don't repeat the word "which" twice.

dash.el
```diff
@@ -2166,8 +2166,42 @@ See `-let' for the description of destructuring mechanism."
...
+           (debug (&define name sexp
+                           [&optional stringp]
+                           [&optional ("declare" &rest sexp)]
+                           [&optional ("interactive" interactive)]
+                           def-body)))
+  (cond
+   ((nlistp match-form)
+    (signal 'wrong-type-argument (list #'listp match-form)))
+   ;; no destructuring, so just return regular defun to make things faster
+   ((-all? 'symbolp match-form)
```
basil-conto, line 2196: Nit: `#'symbolp`

dash.el
```diff
@@ -2166,8 +2166,42 @@ See `-let' for the description of destructuring mechanism."
...
+    `(defun ,name ,match-form ,@body))
+   (t
+    (let* ((inputs (--map-indexed (list it (make-symbol (format "input%d" it-index))) match-form)))
```
basil-conto, line 2199: Nit: This `let*` can be turned into a `let`.

dash.el
basil-conto: Hopefully @Fuco1 will enlighten us.

dash.el
```diff
@@ -2166,8 +2166,42 @@ See `-let' for the description of destructuring mechanism."
...
+      ;; TODO: because inputs to the defun are evaluated only once,
+      ;; -let* need not to create the extra bindings to ensure that.
+      ;; We should find a way to optimize that.  Not critical however.
+      `(defun ,name ,(-map #'cadr inputs)
+         (-let ,inputs ,@body))))))
```
basil-conto, line 2204: This `-let` should be a `-let*` for consistency with `-lambda` and for more flexibility in destructuring.

dash.el
basil-conto: Hopefully @Fuco1 will enlighten us.

dash.el
yyoncho: Pushed new revision with this change reverted. But I wonder will this be really a breaking change? I guess the only way to access the previous param is with creating custom destructuring and with lexical binding off.

basil-conto: Here's a simple way:

```el
(funcall (-lambda ((x) (&plist x y)) y)
         '(:a)
         '(:b b :a a))
```

But I looked again and it seems `-let` is actually implemented in terms of `-let*`, so the breakage I was thinking of doesn't seem to be possible.

yyoncho:
Ping.

ericdallo:
any news on that @basil-conto ?

basil-conto:
> any news on that @basil-conto ?

Any news on what?

I lack both the ability to check whether @yyoncho's copyright assignment is on file, and to merge this PR.

ericdallo:
> I lack both the ability to check whether @yyoncho's copyright assignment is on file, and to merge this PR.

Yeah, sorry for that :/

basil-conto:
> Yeah, sorry for that :/

No worries. :) Perhaps if the `dash.el` maintainers are looking for more help I could use it as justification to ask for access to the copyright assignment list, which would allow synchronising this repository with GNU ELPA, which I already have push rights to.

yyoncho:
FWIW I started copyright assignment but it is not a blocker for the merge either way since I have already 2 commits with more than 15(?) lines in the repo. And generally, we need it in melpa, not sure if having a newer version in elpa will be sufficient.

basil-conto:
> it is not a blocker for the merge either way since I have already 2 commits with more than 15(?) lines in the repo

There was a lot of effort put into getting `dash.el` on GNU ELPA, so it's unfortunate if copyright status hasn't been checked since then.

> And generally, we need it in melpa, not sure if having a newer version in elpa will be sufficient.

The status of `dash.el`, as a widely used Emacs package, on GNU ELPA is far more important than its status on MELPA. The latter is fine for high volumes of community packages, but the former is guaranteed to always be available and maintained by core Emacs developers, which is more important for packages in which code quality and stability is critical.

yyoncho:
> > And generally, we need it in melpa, not sure if having a newer version in elpa will be sufficient.
> 
> The status of `dash.el`, as a widely used Emacs package, on GNU ELPA is far more important than its status on MELPA.

By "we", I meant `emacs-lsp` team. Having it in elpa does not work for us because the melpa version will be higher(I guess).

basil-conto:
> Having it in elpa does not work for us because the melpa version will be higher(I guess).

I don't see how that would be a problem.

yyoncho:
> I don't see how that would be a problem.

package.el will pick the melpa version and we will break if it does not have -defun. We will break even now if anyone was using the elpa version because AFAICT dash was not updated in elpa for like 5 years...

basil-conto:
> package.el will pick the melpa version and we will break if it does not have -defun.

If GNU ELPA has `-defun`, then MELPA will also have it.

> We will break even now if anyone was using the elpa version because AFAICT dash was not updated in elpa for like 5 years...

Which is something that I would like to change.

magnars:
I agree with you, Basil, that we should wait for the CA to be in order, and
I’m happy that you’re putting in the effort to get us back on track.

man. 1. jun. 2020 kl. 22:18 skrev Basil L. Contovounesios <
notifications@github.com>:

magnars:
Since Emacs Lisp has a global namespace, you could include the function in
your own code while waiting for this to propagate to the package repos,
perhaps?

man. 1. jun. 2020 kl. 21:17 skrev Ivan Yonchovski <notifications@github.com

basil-conto:
> Since Emacs Lisp has a global namespace, you could include the function in
your own code while waiting for this to propagate to the package repos,
perhaps?

I haven't tested this, but would `(defalias 'foo (-lambda ...))` effectively do the same as `(-defun foo ...)`?

basil-conto:
> FWIW I started copyright assignment

Any updates on that?

> I haven't tested this, but would `(defalias 'foo (-lambda ...))` effectively do the same as `(-defun foo ...)`?

It seems to - so is `-defun` useful enough to have on its own? Or can we just use `defalias`+`-lambda` instead?

yyoncho:
> > FWIW I started copyright assignment
> 
> Any updates on that?

I will try to finish that these days.

> 
> > I haven't tested this, but would `(defalias 'foo (-lambda ...))` effectively do the same as `(-defun foo ...)`?
> 
> It seems to - so is `-defun` useful enough to have on its own? Or can we just use `defalias`+`-lambda` instead?

I don't insist on getting this in - we already defined lsp-defun in lsp-mode.

PR #339, opened 2020-06-10
Remove duplicate definition.

phst:
‘-partial’ is defined in dash.el and dash-functional.el.  Since the latter
loads the former, define it only in the former.

### DIFF
--- a/dash-functional.el
+++ b/dash-functional.el
@@ -31,13 +31,6 @@
 
 (require 'dash)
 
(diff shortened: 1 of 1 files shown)


phst:
Thanks for the review! Would you mind merging this PR?

basil-conto:
Sorry, I don't have push access.

magnars:
Thanks for the fix, @phst.

PR #342, opened 2020-08-10
Replaced all occurrences of `even?` with `evenp`

cfraizer:
Also replaced occurrences of `odd?` with `oddp`.

Emacs Lisp provides `evenp` and `oddp` and apparently the use of (not defined) functions `even?` and `odd?` caused confusion for some readers. [I won't cry if you ignore this pull request.]

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,7 +363,7 @@ Alias: `-replace-where`
 See also: [`-update-at`](#-update-at-n-func-list)
 
 ```el
-(-map-when 'even? 'square '(1 2 3 4)) ;; => '(1 4 3 16)
+(-map-when 'evenp 'square '(1 2 3 4)) ;; => '(1 4 3 16)
 (--map-when (> it 2) (* it it) '(1 2 3 4)) ;; => '(1 2 9 16)
 (--map-when (= it 2) 17 '(1 2 3 4)) ;; => '(1 17 3 4)
 ```
@@ -375,7 +375,7 @@ Replace first item in `list` satisfying `pred` with result of `rep` called on th
 See also: [`-map-when`](#-map-when-pred-rep-list), [`-replace-first`](#-replace-first-old-new-list)
 
 ```el
-(-map-first 'even? 'square '(1 2 3 4)) ;; => '(1 4 3 4)
+(-map-first 'evenp 'square '(1 2 3 4)) ;; => '(1 4 3 4)
 (--map-first (> it 2) (* it it) '(1 2 3 4)) ;; => '(1 2 9 4)
 (--map-first (= it 2) 17 '(1 2 3 2)) ;; => '(1 17 3 2)
 ```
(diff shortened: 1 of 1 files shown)


basil-conto:
> Emacs Lisp provides `evenp` and `oddp`

Only if you load the `cl.el` library, which is deprecated, at runtime.

> the use of (not defined) functions `even?` and `odd?`

They're defined in the file that generates the examples, `dev/examples.el`.

That said, it's always irked me that the output of this file does not include their definitions.

Instead of using deprecated `cl.el` functions, my suggestion would therefore be to either change the examples so that they use some other built-in functions; use the corresponding `cl-lib.el` definitions; or change the example generation so that the definitions used therein are also published. My personal favourite would be the first option.

basil-conto:
Note also that you shouldn't modify `README.md`, as it is generated from other files, e.g. `readme-template.md` and `dev/examples-to-docs.el`.

magnars:
How about 4) Rename to `-even?` and `-odd?` and include them in dash proper? Not because they are a perfect match for the dash API, but because they are slightly useful and would make running the examples easier?

basil-conto:
Is it really worth further blurring the boundaries of the Dash API for the sake of how and which examples are published?

I'm obviously not going to oppose the decision, I just think there are simpler solutions.

magnars:
It wasn’t a decision. It was a suggestion. As implied by the question
marks. Go for whatever feels best.

tir. 11. aug. 2020 kl. 12:22 skrev Basil L. Contovounesios <
notifications@github.com>:

basil-conto:
> It wasn’t a decision. It was a suggestion. As implied by the question marks.

Sorry, I didn't mean to come across as accusatory. I only meant to say that I'm ultimately fine with any of the solutions, but don't think the generation of examples is worth going into too much trouble for.

cfraizer:
Thanks for all the feedback.

I'll revert the changes to the README.md (of course)

And, if you concur, I'll go with @basil-conto 's first suggestion and alter the examples to use, say `integerp` or `zerop`?

basil-conto:
> And, if you concur, I'll go with @basil-conto 's first suggestion and alter the examples to use, say `integerp` or `zerop`? [I'm open to other suggestions and I'm okay if you suggest something else after seeing them changed.]

I've since been chipping away at the non-standard functions whenever I was changing tests for other reasons, and finally `odd?` started hanging low enough that I just eliminated the handful of cases left in commit ce4a34422945ac65bd2600e13b8c18f21117e32d.

I'm not going to tackle `even?` or `square` very soon, so would you like to rebase this branch off `master` and continue from there?

I think such a cumulative change should be exempt from copyright paperwork, but if you'd like to continue contributing to Dash, Emacs, or other FSF-copyrighted projects, then I'd encourage you to start the copyright assignment process, if you haven't already :).  If you're unfamiliar with this, see [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html).  Thanks.

basil-conto:
> > Emacs Lisp provides `evenp` and `oddp`
> 
> Only if you load the `cl.el` library, which is deprecated, at runtime.

Update: Emacs 31 now provides `evenp` and `oddp` OOTB.

My personal preference remains to limit examples in Dash docs to using the `dash.el` API and vanilla Elisp that works in all supported Emacs versions.

Hopefully someone would still like to make this happen (@cfraizer ?) :). Otherwise I'll get to it at some point™...

PR #347, opened 2020-11-06
Add `-defun'

nbfalcon:
`-defun` is like `-lambda', but with destructuring. To implement it, abstract
the arglist and form generation code of `-lambda` into separate functions.

Additional benefits of the refactoring are:

- `-lambda` (and by extension, `-defun`) can now accept empty argument lists
- `-lambda` (and `-defun`) no longer rebinds symbols (fixing the TODO)

Various declaration forms should be handled properly in `-defun`.

Motivation: when this is merged, `lsp-defun' can be made an alias for `-defun`.

- Please note the use of `intern` instead of `make-symbol`. Using the latter instead caused test-failures.
- What `git-commit-major-mode` do you prefer? I assumed and used `git-commit-elisp-text-mode` here

I'll add examples soon.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -2136,6 +2136,66 @@ because we need to support improper list binding."
       `(let ,inputs
          (-let* ,new-varlist ,@body)))))
 
+(defun dash--make-arglist (args)
+  "Make ARGS a function arglist for `dash--destructure-arglist'."
+  (--map-indexed
+   ;; No need to destructure a symbol to itself.
+   (if (symbolp it) it (intern (format "input%d" it-index))) args))
+
+(defun dash--progn (body-forms)
+  "Wrap BODY-FORMS in a `progn'.
+The `progn' is omitted if it isn't needed. This function is just
+`macroexp-progn' reimplemented."
+  (if (cdr body-forms)
+      `(progn ,@body-forms)
+    (car body-forms)))
+
+(defun dash--destructure-arglist (args body-forms)
+  "Destructure function arguments ARGS using `-let'.
+The result, an ELisp-form, is valid only in a function whose
+arguments are the result of transforming ARGS with
+`dash--make-arglist'. In the resulting form, BODY-FORMS are
+executed within the bound environment."
+  (unless (listp args)
+    (signal 'wrong-type-argument "match-form must be a list"))
+  (let ((let-bindings
+         (-remove #'null
+                  (--map-indexed
+                   ;; Symbols shouldn't be rebound; they can be taken from the
+                   ;; surrounding environment directly.
+                   (unless (symbolp it)
+                     (list it (intern (format "input%d" it-index))))
+                   args))))
+    (if let-bindings
+        `(-let ,let-bindings ,@body-forms)
+      (dash--progn body-forms))))
+
+(defun dash--decompose-defun-body (body)
+  "Split `defun'-body BODY into declarations and the real body.
+The body of a `defun' can start with various declare forms like
+interactive, `declare', ... and a docstring. Split BODY into a
+list (DECLS REALBODY)."
+  (let* ((docstring? (car body))
+         (result (--split-with (memq (car-safe it) '(declare interactive))
+                               (if (stringp docstring?) (cdr body) body))))
+    (when (stringp docstring?)
+      (push docstring? (car result)))
+    result))
+
+(defmacro -defun (name match-form &rest body)
+  "Like `-lambda', but as a `defun'.
+Define a function called NAME with destructuring.
+
+MATCH-FORM is like in `-lambda'. Except for the (optional)
+additional destructuring, this function behaves exactly like
+`defun' (in terms of `declare', ...).
+
+\(-defun NAME MATCH-FORMS &optional DOCSTRING DECL &rest BODY)"
+  (-let [(decls realbody) (dash--decompose-defun-body body)]
+    `(defun ,name ,(dash--make-arglist match-form)
+       ,@decls
+       ,(dash--destructure-arglist match-form realbody))))
+
 (defmacro -lambda (match-form &rest body)
   "Return a lambda which destructures its input as MATCH-FORM and executes BODY.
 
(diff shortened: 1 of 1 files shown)


nbfalcon:
I haven't noticed yyoncho's PR... This is going to be awkward.

nbfalcon:
I believe my PR to be better: it has no code duplication, and it doesn't rebind symbols, so things like (-lamba (foo)) are the same as (lambda (foo)). Is signing a copyright disclaimer enough, and if so, how would I go about doing that?

nbfalcon:
I'll add support for &optional.

basil-conto:
> Is signing a copyright disclaimer enough, and if so, how would I go about doing that?

Just email the completed form at https://git.savannah.gnu.org/cgit/gnulib.git/tree/doc/Copyright/request-assign.future to assign@gnu.org; they'll give you further instructions. It's generally a matter of scanning and emailing a scanned document, or in some countries digitally signing a document. See [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html) for more.

dev/examples.el
```diff
@@ -1259,6 +1259,17 @@ new list."
...
+  (defexamples -defun
+    (progn (-defun example/cdr ((_ . tail)) tail)
+           (example/cdr '(a . b))) => 'b
+    (progn (-defun example/car ((cur)) cur)
+           (example/car '(a . b))) => 'a
+    (progn (-defun example/add-cons ((a . b))
+             "Add the `car' and `cdr' of INPUT0."
+             (interactive (list (cons 1 2)))
+             (+ a b))
+           (command-execute #'example/add-cons)) => 3)
```
basil-conto, line 1271: Aren't examples limited to a single line? Have you tried regenerating the docs from this?

basil-conto (commented):
Just an initial comment for now; I'll have a closer look later.

dev/examples.el
nbfalcon: I haven't yet tried regenerating the docs - I'll do that later. I see some two-line examples in `-setq`.

nbfalcon:
Now we also support &rest and &optional.

nbfalcon:
TODO: add &key, making `-defun` a 100% drop-in replacement for `cl-defun`.

nbfalcon:
Now we have -defmacro, and correct function signatures :).

Also, do you think I should implicitly add a signature doc to -lambda, if a docstring is provided? What we have currently:
```emacs-lisp
(-lambda (a b)) = (lambda (a b))
(-lambda (a b) "Some docstring...") = (lambda (a b) "Some docstring...") ; no destructuring and signature comment necessary, as we only use symbols in the arglist
(-lambda ((a b)) "") = (-lambda ((a b)) "\n\n(fn ((a b)))") ; Note the auto-generated signature docstring fragment
(-lambda ((a b))) = (-lambda ((a b))) ; no signature docstring, as the user probably doesn't care if they didn't bother to provide a docstring
```

What I could do instead is to never touch `-lambda` docstrings at all. What do you think?

Other than &key (which may or may not be needed), this PR is ready.

nbfalcon:
I couldn't write a proper edebug-form-spec for -defun, -lambda, -defmacro's lambda-lists, because I still don't understand them.

nbfalcon:
Now I added:
- `(-lambda [sym &as &DocumentSymbol :kind])` = `(-lambda ((sym &as &DocumentSymbol :kind)))`
- `(-lambda ((sym &as &DocumentSymbol :kind))` = `(lambda (sym) (-let (((&DocumentSymbol :kind) sym)))` (&as are destructured directly, without going trough input<n>)

dash.el
```diff
@@ -2209,6 +2209,11 @@ docstring if none is provided."
          `((-let ,let-bindings ,@body))
        body))))
 
+(defun dash--normalize-arglist (arglist)
+  "Make ARGLIST have the form (MATCHERS...).
+If it is a vector, convert it to a single-matcher arglist."
+  (if (vectorp arglist) (list (append arglist nil)) arglist))
```
nbfalcon, line 2215: Can I use `cl-coerce` here? `append` nil seems fragile.

basil-conto: `(append x ())` is the canonical way to convert an array into a list, so I wouldn't worry about it. Also, `dash.el` tries to avoid relying on `cl-lib`.

nbfalcon:
My changes seem to not have broken anything: I am using my modified dash.el with lsp-mode, a heavy user of `-lambda`, without any issues. I have recompiled the former a few times already.

basil-conto:
Thanks, I'll have a closer look in the coming days. Let us know how the copyright assignment process goes.

nbfalcon:
I have sent the E-Mail for the copyright assignment.

dash.el
```diff
@@ -2136,6 +2136,66 @@ because we need to support improper list binding."
...
+  (let ((let-bindings
+         (-remove #'null
+                  (--map-indexed
+                   ;; Symbols shouldn't be rebound; they can be taken from the
+                   ;; surrounding environment directly.
+                   (unless (symbolp it)
+                     (list it (intern (format "input%d" it-index))))
+                   args))))
+    (if let-bindings
+        `(-let ,let-bindings ,@body-forms)
```
nbfalcon: Note the `-let`: previously, `-let*` was used instead, which might be more efficient as the latter is implemented in terms of the latter. Not doing that might also make my various symbol and &as optimizations for naught. What do you think?

basil-conto: I don't think I can say something intelligent on this, as I have not yet looked at the code closely.

But re: `-let` being implemented in terms of `-let*` - I think that's an implementation detail that we should not base other design decisions on, since `-let` would be implemented only in terms of itself.

Fuco1: You need to make sure that input arguments are evaluated in parallel, whether with `-let` or `-let*` doesn't matter much.

Fuco1: You would need to construct some kind of graph of dependencies between the bindings.  Which I suppose was too much work and I simply leveraged the built-in let to handle it for me.  But yes this is a possible optimization.  In practice probably insignificant :)  But it could be intelectually stimulating :D

You're right about the lambda, there the parallel eval is handled by the runtime when the arguments are passed to the "wrapper lambda".

Issue #330, opened 2020-02-09
[Feature request] Clojure's cond-> 


yyoncho:
Example from the clojure docs

``` clojure
(cond-> 1          ; we start with 1
    true inc       ; the condition is true so (inc 1) => 2
    false (* 42)   ; the condition is false so the operation is skipped
    (= 2 2) (* 3)) ; (= 2 2) is true so (* 2 3) => 6 
```

kinderferraz:
Hi

Is it still open for contribution?
I've spent the last few days implementing this macro as an exercise in emacs-lisp, and I'm finally happy with the solution.
I'd like to submit my code for discussion, as a contribution.
However, it would be my first time making a PR, so some guidance is appreciated.

magnars:
Hi @kinderferraz . Yes, `cond->` and `cond->>` would be welcome additions. 👍

PR #349, opened 2020-12-06
Implement -cond-> and -cond->> macros.

kinderferraz:
Hi.

This is the implementation I mentioned in #330.
My main difficulty when porting this from Clojure's definition was regarding `make-symbol` vs. `intern`.
When using make-symbol, I ran into the same problem described [here](https://emacs.stackexchange.com/questions/31253/why-does-my-function-generating-macro-only-work-when-i-evaluate-its-expansion)/ that is, when calling the macro, only the last expression would be tested and run.
Therefore I decided for `intern`. In my tests it did not interfere with other variables, but more tests are required.

Hope this code fits with the package.
Also, some feedback would be appreciated, since this is my first PR.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1591,6 +1591,58 @@ and when that result is non-nil, through the next form, etc."
                    (--> ,result ,form))
                  ,@more))))
 
+(defmacro -cond-> (x &rest clauses)
+  "Conditionally thread X through CLAUSES.
+Threads x (via `->') through each form for which the
+corresponding test expression is true.  Note that, unlike cond
+branching, `-cond->' threading does not short circuit after the
+first true test expression.
+Given some elisp details, current value is exposed as the symbol
+`it'."
+  (declare (debug (form body))
+           (indent 1))
+  (when (-> clauses length (% 2) (= 1))
+    (error "Wrong number of arguments."))
+  (-let* ((it (intern "it"))
+          (steps (-map
+                  (-lambda ((test step))
+                    `(if ,test
+                         (-> it ,step)
+                       it))
+                  (-partition 2 clauses))))
+    `(-let* ((it ,x)
+            ,@(-zip-lists (-cycle '(it))
+                          (butlast steps)))
+       ,@(if (null steps)
+            it
+           (last steps)))))
+
+(defmacro -cond->> (x &rest clauses)
+  "Conditionally thread X through CLAUSES.
+Threads x (via `->>') through each form for which the
+corresponding test expression is true.  Note that, unlike cond
+branching, `-cond->>' threading does not short circuit after the
+first true test expression.
+Given some elisp details, current value is exposed as the symbol
+`it'."
+  (declare (debug (form body))
+           (indent 1))
+  (when (-> clauses length (% 2) (= 1))
+    (error "Wrong number of arguments."))
+  (-let* ((it (intern "it"))
+          (steps (-map
+                  (-lambda ((test step))
+                    `(if ,test
+                         (->> ,it ,step)
+                       ,it))
+                  (-partition 2 clauses))))
+    `(-let* ((,it ,x)
+             ,@(-zip-lists (-cycle '(it))
+                           (butlast steps)))
+       ,@(if (null steps)
+            it
+           (last steps)))))
+
 (defun -grade-up (comparator list)
   "Grade elements of LIST using COMPARATOR relation, yielding a
 permutation vector such that applying this permutation to LIST


magnars:
Hi!

First of all, thanks for your work on this. :) 

There seems to be a long-standing issue with the CI-tests, which is why the checks are failing. I haven't been active in maintaining this repo for quite some time. Sorry about that.

Now, I have a few comments:

- please add some examples in dev/examples.el
- the README is actually generated from the docstring and the examples, so no need to update that manually.

Also, I would prefer if `-cond->` had the same semantics as `cond->` in clojure.  Fewer surprises that way. From what I could tell from the examples in your README, it differs in two major ways:

- the `it` binding
- a default clause with no predicate in front. 

I might have read that last one too quickly, so do correct me if I am wrong.

I haven't enough Emacs Lisp knowledge to know if `intern` is a good substitute for `make-symbol`. Maybe @Fuco1 has some ideas?

Cheers!

kinderferraz:
Hi,
you're welcome.
Thank you for answering.

> please add some examples in dev/examples.el
> the README is actually generated from the docstring and the examples, so no need to update that manually.
> You're welcome :)

I wasn't aware of these documenting features, dev/examples.el is updated now.

As for the differences from Clojure's definitions, 
> - a default clause with no predicate in front
>
> I might have read that last one too quickly, so do correct me if I am wrong.

Indeed, there was a mistake in my examples, it must be correct now.

> - the it binding

This came as a surprise when i was testing. I tested a variable with the same name as the internally bound one, trying to see if there was a bug or an error, and this showed up, so I changed the symbols to `it` as to match other macros. I'm afraid I can't offer a solution to this.

> I haven't enough Emacs Lisp knowledge to know if intern is a good substitute for make-symbol.

Just tested a version with `make-symbol`, and everything worked fine...
had to remove a few commas, though, but they shouldn't be there in the first place.

(Also, I've completely mistyped last commit's message, sorry 
hahahahaha)

kinderferraz:
Should the generated README file be committed as well?

yyoncho:
Using `it` is not desired here since it will unexpectedly shadow the `it` variable which is commonly used in other `dash` macros. Introducing `it` should happen in `--cond->`. You may generate a symbol following what `dash--match-make-source-symbol` do.

kinderferraz:
I'll look into it! :)

Just out of curiosity, I tested some pieces of code, it would appear that these macros don't overshadow each other. Some of these lines were made complex on purpose.
``` emacs-lisp
(-cond->> '(10 20 30 40 50 60)
  (listp it) (--map (/ it 2))
  (numberp it) (--filter (--> it (% it 2) (= it 0)))
  (numberp (car it)) (--reduce (format "%s-%d" acc it))) ;; => "5-10-15-20-25-30"

(--map (-cond-> it
         (numberp it) (/ 2)
         (odd? it) (number-to-string))
       '(10 20 30 40 50)) ;; => '("5" 10 "15" 20 "25")

(--> "abc"
     (-cond-> it
       t (concat "def" "ghi")
       (string= it "abc") it
       (string= it "abcdefghi") (split-string "" t)) ;; =>  '("a" "b" "c" "d" "e" "f" "g" "h" "i")
     (--reduce-from (concat acc it) "" it)) ;; => "abcdefghi"
```

Unfortunately, this is no exhaustive test of all available macros, and breaks uniformity of available macros.

kinderferraz:
Just committed using `dash--match-make-source-symbol`, with updated examples.
There should be no way to access current value in tests, and unless there's a variable named `it` outside the macro, there should be an error. :)

I guess this leaves for two options for anaphoric versions of this macro, one with `-->` macro, which would prevent accessing current values in tests, i.e:
```el
(-cond--> '(10 20)
    t (--map (/ 2) it)
    (= 1 1) (setf (car it) 11) ;; => '(11 10)
```
and another following the last attempt, exposing the symbol `it`. Let me know which one is best! :)

magnars:
I don't think `it` makes any sense in `cond->`, but if I'm wrong it should certainly be named `--cond->` and `--cond->>`, to match the other anaphoric macros.

kinderferraz:
I see.
You'll find current committed code to behave as you expect, and strictly follows Clojure's semantics.

magnars:
I'm glad to hear that. :) It seems the example generating code needs to be run another time, since the examples in the readme are outdated, but otherwise this looks good to me.

magnars:
Well done. In order for this code to be allowed into the ELPA repo, all contributors must sign a copyright assignment. 

basil-conto said: 

> Just email the completed form at https://git.savannah.gnu.org/cgit/gnulib.git/tree/doc/Copyright/request-assign.future to assign@gnu.org; they'll give you further instructions. It's generally a matter of scanning and emailing a scanned document, or in some countries digitally signing a document. See (info "(emacs) Copyright Assignment") for more.

kinderferraz:
Sent! 😄
I'll leave a message here when the process is completed.

README.md
```diff
@@ -1,8 +1,4 @@
-![GitHub Workflow Status](https://img.shields.io/github/workflow/status/magnars/dash.el/CI)
-[![MELPA](https://melpa.org/packages/dash-badge.svg)](https://melpa.org/#/dash)
-[![MELPA Stable](https://stable.melpa.org/packages/dash-badge.svg)](https://stable.melpa.org/#/dash)
-
-# <img align="right" src="https://raw.github.com/magnars/dash.el/master/rainbow-dash.png"> dash.el
+# <img align="right" src="https://raw.github.com/magnars/dash.el/master/rainbow-dash.png"> dash.el ![GitHub Workflow Status](https://img.shields.io/github/workflow/status/magnars/dash.el/CI)
```
basil-conto, line 1: Could you please re-run `create-docs.sh`? I'd forgotten about `readme-template.md` when I merged the last PR that added these badges. Some unrelated changes below should also be covered by that.

dash.el
```diff
@@ -1591,6 +1591,54 @@ and when that result is non-nil, through the next form, etc."
                    (--> ,result ,form))
                  ,@more))))
 
+(defmacro -cond-> (x &rest clauses)
+  "Conditionally thread X through CLAUSES.
+Threads x (via `->') through each form for which the
+corresponding test expression is true.  Note that, unlike cond
+branching, `-cond->' threading does not short circuit after the
+first true test expression."
```
basil-conto, line 1599: Some nits/suggestions:
```
x -> X
true -> non-nil
cond -> `cond'
branching -> clauses
```
Maybe it's just my unfamiliarity with Clojure, but it's not clear to me what "test expression" is referring to. I would appreciate an explanation of how it relates to `CLAUSES`, which is mentioned in the beginning but not subsequently explained. I think some destructuring like in the docstring of `cond` would be helpful, as would mentioning the expected length of `CLAUSES`.

dash.el
```diff
@@ -1591,6 +1591,54 @@ and when that result is non-nil, through the next form, etc."
...
+  (declare (debug (form body))
+           (indent 1))
+  (when (-> clauses length (% 2) (= 1))
```
basil-conto, line 1602: This is checking that the number of clauses is odd, right? Shouldn't it be even instead?

[Not that my vote means much, but I'd vote for writing in simple Elisp rather than Clojure where the latter doesn't significantly simplify things, such as in this one-liner.]

dash.el
```diff
@@ -1591,6 +1591,54 @@ and when that result is non-nil, through the next form, etc."
...
+    (error "Wrong number of arguments"))
```
basil-conto, line 1603: There's a standard `wrong-number-of-arguments` for this; see [`(info "(elisp) Standard Errors")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Standard-Errors.html).

dash.el
```diff
@@ -1591,6 +1591,54 @@ and when that result is non-nil, through the next form, etc."
...
+  (-let* ((g (dash--match-make-source-symbol))
```
basil-conto, line 1604: Wouldn't a simple `make-symbol` with a more descriptive name do? We don't strictly need monotonically increasing symbol names since this is a single binding.

dash.el
```diff
@@ -1591,6 +1591,54 @@ and when that result is non-nil, through the next form, etc."
...
+                    `(if ,test
+                         (-> ,g ,step)
+                       ,g))
+                  (-partition 2 clauses))))
+    `(-let* ((,g ,x)
+             ,@(-zip-lists (-cycle (list g))
+                           (butlast steps)))
+       ,@(if (null steps)
+             g
+           (last steps)))))
```
basil-conto, line 1616: If you build up `steps` in reverse, e.g. with a double `pop` and a `push` per clause, then you could use `car` and `cdr` instead of `last` and `butlast`, right?

dash.el
```diff
@@ -1591,6 +1591,54 @@ and when that result is non-nil, through the next form, etc."
...
+(defmacro -cond->> (x &rest clauses)
+  "Conditionally thread X through CLAUSES.
+Threads x (via `->>') through each form for which the
+corresponding test expression is true.  Note that, unlike cond
+branching, `-cond->>' threading does not short circuit after the
+first true test expression."
+  (declare (debug (form body))
+           (indent 1))
+  (when (-> clauses length (% 2) (= 1))
+    (error "Wrong number of arguments."))
```
basil-conto, line 1627: Same comments as with `-cond->`, but also error messages shouldn't end in a full stop; see [`(info "(elisp) Error Symbols")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Error-Symbols.html).

The two macros also seem to be identical, bar the threading macro used. Can't we define them in terms of the same function for less duplication?

dev/examples.el
```diff
@@ -1010,7 +1010,32 @@ new list."
...
+  (defexamples -cond->>
+    (let ((string "abc"))
+      (-cond->> "abc"
+      (stringp string) (concat "def" "ghi"))) => "defghiabc"
+    (-cond->> '(1 2 3)
+      nil (+ 10 100 1000 )
+      t   (-filter 'even?)) => '(2)
+    (-cond->> "ghi"
+      (= 1 1) (concat "abc" "def")
+      (= 0 1) (split-string)) => "abcdefghi"))
```
basil-conto, line 1038: The first three examples are automatically added to the `README.md` as a single line unfortunately, so I'd recommend against overly long multiline examples.  Examples from the fourth one onwards are still run as unit tests, so you can add longer examples there, but the first three should be short and sweet.

dev/examples.el
```diff
@@ -1010,7 +1010,32 @@ new list."
...
+    (let ((list '(:name "John" :age 24)))
+      (-cond-> list
+        (not (plist-get list :name)) (plist-put :name "Owen")
+        (not (plist-get list :age)) (plist-put :age 40)
+        (not (plist-get list :address)) (plist-put :address "123, Saint St.")))
+    => '(:name "John" :age 24 :address "123, Saint St."))
+
+  (defexamples -cond->>
+    (let ((string "abc"))
+      (-cond->> "abc"
```
basil-conto, line 1031: Shouldn't this use the variable `string` instead of the same string literal?

dev/examples.el
```diff
@@ -1010,7 +1010,32 @@ new list."
...
+      (stringp string) (concat "def" "ghi"))) => "defghiabc"
```
basil-conto, line 1032: Indentation seems off here.

dev/examples.el
```diff
@@ -1010,7 +1010,32 @@ new list."
...
+    (-cond->> '(1 2 3)
+      nil (+ 10 100 1000 )
```
basil-conto, line 1034: Trailing space.

dev/examples.el
```diff
@@ -1010,7 +1010,32 @@ new list."
...
+      t   (-filter 'even?)) => '(2)
```
basil-conto, line 1035: Nit: `'` -> `#'` for function symbols.

dev/examples.el
basil-conto, line 1038: These conditions are the same as `t` and `nil` without demonstrating anything new about `-cond->>`. It would be nice to showcase something more realistic.

basil-conto (commented):
Thanks for working on this! See some comments below.

basil-conto:
> My main difficulty when porting this from Clojure's definition was regarding `make-symbol` vs. `intern`.

All symbols that you normally refer to while using Emacs, such as the `setq`, `var`, and `nil` in the form `(setq var nil)` are interned, i.e. they are stored in an obarray. Obarrays can be thought of as weird hash tables, where interning a symbol is the same as inserting it into a bucket in the obarray that is determined by hashing the symbol's name. Emacs stores variable, function, and keyword symbols, etc. in a global obarray called `obarray`, but you can also create your own obarrays and intern symbols in them.

Since interning symbols by default happens in the global `obarray`, macros that create local variable bindings have to be careful to use symbols that are guaranteed to not clash with the forms that are passed to those macros as arguments. Enter `make-symbol`: it returns a symbol object (with the usual value, function, property list slots, etc.) that is crucially not interned in any other obarray, so binding it to different values will not overwrite any other variables in the program.

See more at [`(info "(elisp) Symbols")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Symbols.html), specifically under [`(info "(elisp) Creating Symbols")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Creating-Symbols.html).

kinderferraz:
@basil-conto 
Thank you so much for this review!!
I'll read it all and start working on this asap.

README.md
kinderferraz: I've update documentation and examples, and tried to run `create-docs.sh`. It worked alright, except for `dash.info`. Most of the file seems to be altered and changes feel unnecessary, like using double spaces after periods and using
`’function’` instead of ``` `function' ```
I've tried updating makeinfo, checked PATH in case it points to an older version,  but I don't know what to do.

dash.el
kinderferraz: It does check if the number of clauses is odd, in which case there is an error, i.e there is a test with no expression or vice-versa. I guess there is no problem with using `(unless ...)`.

[This one liner maybe a vice of mine hahaha]

dash.el
kinderferraz: Sorry about the full stop, never making this mistake again! 😄

The idea of defining a common structure has crossed my mind a few times, but since I set up to port these macros, i didn't pay much attention to it.

dash.el
basil-conto: Oops, got mixed up. I wasn't suggesting switching the `when` to `unless`, FTR.

README.md
basil-conto: You could just leave out the changes to the manual from this PR and I or someone else will update them after the fact.

[I'd like to move away from checking in the generated Info manual. I don't know why it was added to begin with, but if it was related to GNU ELPA's tooling then that should have been recently fixed.]

kinderferraz: Very well! I'll focus on the code and examples. 😄

dev/examples.el
kinderferraz: Sorry.

dev/examples.el
kinderferraz: I've removed examples such as this :)
I hope the new ones are better.

dash.el
kinderferraz: Perhaps. I was sure using a static symbol would expose it, bound to the the value being passed between expressions.
(This behaviour is clearly undesired, but I wished for it in the first versions. It made sense to test the current value, not just a snapshot of the first one. This is why i looked for `intern` (and became more confused).)
A quick test made me realise I was wrong. It will be corrected soon!

dash.el
kinderferraz: Sorry, I don't understand what you mean here. :(

kinderferraz: Perhaps I should make myself more clear:
I do understand the the general idea, how to use the built-in functions and why using them is a good idea.
What I'm not sure is the sequence of steps needed to achieve this.

basil-conto: Does the following make sense?

```el
(defun dash--cond-thread (name thread expr clauses)
  "Implement macroexpansion of NAME, one of `-cond->' or `-cond->>'.
THREAD is the corresponding threading macro, one of `->' or
`->>'.  EXPR and CLAUSES are as in `-cond->'.  Signal
`wrong-number-of-arguments' if the number of CLAUSES is odd."
  (cond
   ((null clauses) expr)
   ((let ((len (length clauses)))
      (when (= (% len 2) 1)
        (signal 'wrong-number-of-arguments (list name (1+ len))))))
   ((let ((g (make-symbol "g"))
          binds)
      (while clauses
        (let ((test (pop clauses))
              (form (pop clauses)))
          (push `(,g (if ,test (,thread ,g ,form) ,g)) binds)))
      `(let* ((,g ,expr)
              ,@(nreverse (cdr binds)))
         ,(car (cdar binds)))))))
```

Then `-cond->` and `-cond->>` can be implemented as follows:

```el
(defmacro -cond-> (expr &rest clauses)
  (declare (indent 1))
  (dash--cond-thread #'-cond-> #'-> expr clauses))

(defmacro -cond->> (expr &rest clauses)
  (declare (indent 1))
  (dash--cond-thread #'-cond->> #'->> expr clauses))
```

Key differences:
- If there are no clauses, just return the first argument instead of signalling an error.
- `signal` takes two arguments.

kinderferraz: It's amazing!! I wish I had thought of this.

basil-conto: If you think my suggestion is useful and clear enough, then feel free to incorporate it in your own code.

kinderferraz: Thank you!
I'm incorporating the part where steps are built whit two pop and a push, and the key differences listed above. But if I incorporated all the code, I'd be taking credit for your work. Perhaps you could update over this PR eventually.

basil-conto: > But if I incorporated all the code, I'd be taking credit for your work.

But my work is already built on top of your work. ;) Also, my public suggestions are not copyrighted and I don't care about attribution.

> Perhaps you could update over this PR eventually.

Sure, if that's what you prefer, I'd be happy to oblige.

kinderferraz: > But my work is already built on top of your work. ;)

That's a nice way of looking at it.

> Sure, if that's what you prefer, I'd be happy to oblige.

IMO, this is the best option :)

basil-conto: Cool, then I think we're just waiting for your CA process to complete, and we should be good to go.

BTW, if you rebase this branch off the latest `master` then the CI tests should stop wetting the bed. :)

dash.el
Fuco1: `make-symbol` always creates a new symbol, even if it has the same name, they are in a different "intern table":

    (eq (make-symbol "a") (make-symbol "a"))  ;; => nil

`intern` puts the symbol in the global intern table *if it doesn't exist*, otherwise returns an existing symbol:

    (eq (intern "a") (intern "a"))  ;; => t

So that's it.  If you need a unique symbol which doesn't exist anywhere in emacs, use `make-symbol`, if you need to refer to a symbol which could've possible been created elsewhere, use `intern`:

    (funcall (intern "concat") "a" "b")  ;; => "ab", we refer to the concat function built-in in emacs
    (funcall (make-symbol "concat") "a" "b") ;; => error (void-function concat) because this symbol is looked up in it's own "universe", not the global one.

PR #328, opened 2019-12-03
Use github actions instead of travis

conao3:
Hi! Your travis CI with something wrong Emacs-26.
I suggest switch GitHub Actions using purcell's [setup-emacs](https://github.com/purcell/setup-emacs) from TravisCI.

I have checked the workflow works well in my fork ([log](https://github.com/conao3/dash.el/commit/d38088ffdcf48696670ec11d113451b50c5089d7/checks?check_suite_id=338889566))

PR #351, opened 2020-12-15
Update CI configuration

basil-conto:
Remove Travis CI + Cask + EVM which are slow and problematic.

* `.github/workflows/test.yml`: Run CI also for PRs and with the latest Emacs release and Checkout Action versions.
* `.travis.yml`:
* `run-travis-ci.sh`: Remove files.

Re: #328.
Cc: @conao3

---

**Questions**:
1. Who gets notified when CI fails, and how?
2. Tests currently always succeed with Emacs Snapshot, even when there are failures. Can we realistically disallow failures with Emacs Snapshot, or will that be too noisy?


conao3:
I don't know what I should comment on, but this change looks good to me.

basil-conto:
> I don't know what I should comment on, but this change looks good to me.

That's all I needed to hear, thanks. :)

Fuco1:
We should keep the cask file in case some people use the cask link feature.  I'm still sad that Elisp has no "package" definition file, and with cask falling out of favor we might actually go backwards when it comes to package development :/

Fuco1 (changes requested):
let's keep the cask file

basil-conto:
> We should keep the cask file

Sure, no problem.

> in case people use the cask link feature

What's that?

> I'm still sad that Elisp has no "package" definition file

What do you mean by that? How is it different to ELPA (the protocol's) `<foo>-pkg.el` files?

Any problems we have with Emacs packaging should be reported and discussed upstream, so that Emacs can improve for everyone.  Maybe it's my unfamiliarity with Cask talking, but I don't see why third-party solutions should be desirable here.

Fuco1:
> What do you mean by that? How is it different to ELPA (the protocol's) <foo>-pkg.el files?

well of course you're right.  I just never used this directly and so I forgot it even existed.

Cask link is a feature for local development where you can link to a folder instead of a repository (remote or otherwise).  It requires that both the source and target packages have a cask file.

Fuco1 (approved):

basil-conto:
> Cask link is a feature for local development where you can link to a folder instead of a repository (remote or otherwise). It requires that both the source and target packages have a cask file.

Looks like Cask linking also works with standard `-pkg.el` files; see https://cask.readthedocs.io/en/latest/guide/usage.html#cask-link. Does that address your concerns?

basil-conto:
> > Cask link is a feature for local development where you can link to a folder instead of a repository (remote or otherwise). It requires that both the source and target packages have a cask file.
> 
> Looks like Cask linking also works with standard `-pkg.el` files; see https://cask.readthedocs.io/en/latest/guide/usage.html#cask-link. Does that address your concerns?

@Fuco1 Ping.

PR #354, opened 2020-12-30
Improve take/drop definitions

basil-conto:
The main changes here are:
1. Consistently return a copy from take/drop functions.
2. Optimise `-take`.

The reason I'm submitting this PR is because I'm not sure how controversial (1) is.  Is it safe to change `--drop-while`, `-drop`, and `-drop-last` to always return a copy, for consistency with the rest of the Dash API? Or is that too much of a breaking change?

---

* `dash.el` (`--take-while`): Use `push`.
(`-take-while`, `-drop-while`, `-take-last`): Improve docstring.
(`--drop-while`, `-drop`, `-drop-last`): Return a copy.
(`-take`): Optimize in terms of `--take-while`.

* `dev/examples.el` (`-take`, `-take-last`, `-drop`, `-drop-last`, `-take-while`, `-drop-while`): Add more tests.

* `README.md`:
* `dash.info`:


magnars:
I certainly hope people are not using this bug as a feature. Is there any
way we could find out? There are so many packages using dash. It would be a
shame to break people’s editors, but I would also like for this bug to be
fixed.

ons. 30. des. 2020 kl. 05:30 skrev Basil L. Contovounesios <
notifications@github.com>:

basil-conto:
> Is there any way we could find out?

Apart from waiting for bug reports to trickle in, I don't think so. ;)

> It would be a shame to break people’s editors, but I would also like for this bug to be fixed.

Here are the options I can think of:
1. Bite the bullet, and if anyone *does* end up complaining, tell them that their code relied on an oversight that's easy to fix on their end.
2. Put off this change until Dash v3, which I understand intends to introduce other breaking changes as well.
3. Add new, fixed versions of `--drop-while`, `-drop`, and `-drop-last` with different names e.g. with a `-safe` suffix.
4. Give `--drop-while`, `-drop`, and `-drop-last` an optional argument telling them whether to make a copy.

Preferences?

magnars:
I say we hope this bug has not been someone’s feature, and go for #1. I am
not a big fan of a breaking version 3 of dash, so I hope we can avoid that.

ons. 30. des. 2020 kl. 16:31 skrev Basil L. Contovounesios <
notifications@github.com>:

basil-conto:
> I say we hope this bug has not been someone’s feature, and go for #1.

WFM.  I'll wait a little longer in case someone else objects before merging this PR.

> I am not a big fan of a breaking version 3 of dash, so I hope we can avoid that.

I agree.

Fuco1:
Thanks for spotting this.  I certainly think we should maintain the promise that these functions are pure and I think dash pushed this premise quite a lot when it comes to expectations.  

I'm personally a little less breaking-changes avoiding type than magnars (which is good so we balance each other out) so I would go ahead with this even in case of breaking some code.

Maybe we can also add explicitly somewhere a warning that all dash functions are expected to be used in a pure way except those which are explicitly marked as not pure, such as `!cdr` and such.

basil-conto:
I just noticed an issue with this purity while simplifying `-split-at`: copying a list is only possible if it's a proper list, i.e. not circular or dotted.

This means that functions calling `copy-sequence`, `append` et al. for purity reasons can no longer accept e.g. the result of `-cycle`, as seen in `dev/examples.el`:

```el
(-map (-partial '-take 5) (-split-at 5 (-cycle '(1 2 3))))
;; -split-at: List contains a loop: (2 3 1 2 3 . #2)
```

So, do we prioritise purity or flexibility?  Repeating some of the aforementioned options:
1. Make all functions pure and document them as such at a loss of working with improper lists.  This is also a slightly breaking change.
2. Document which functions are not entirely pure and be done with it.
3. Define additional pure versions of the currently impure functions.
4. Give currently impure functions an optional argument that toggles their purity.

BTW, ["pure" in Elisp](https://www.gnu.org/software/emacs/manual/html_node/elisp/What-Is-a-Function.html) just means returning the same stateless result based on the arguments. So returning the same list or one of its links still counts as pure, it just means the caller has to be more careful about what they do with the result. For example, they shouldn't destructively modify a cons that's part of a literal list constant [`(info "(elisp) Mutability")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Mutability.html).

I'm therefore leaning towards (2), which might involve reverting some of these `copy-sequence` additions. Thoughts?

Fuco1:
What I would expect is that if I do `(setq y (-dash-function x))` and than mutate `y` by side-effect, this change won't affect `x` (the input).  Otherwise it really is a very weak guarantee.

> I just noticed an issue with this purity while simplifying -split-at: copying a list is only possible if it's a proper list, i.e. not circular or dotted.

Why would this be the case?  Any structure can be walked and copied.

basil-conto:
> Otherwise it really is a very weak guarantee.

Sure, but that's the case in Elisp anyway, and long-standing Dash behaviour in the functions being discussed.

> Why would this be the case? Any structure can be walked and copied.

How do you realistically copy the result of `(-cycle '(1 2 3))`?

basil-conto:
Thinking about this a bit more: following this PR, the following functions call `copy-sequence` on the tail of the given list: `--drop-while`, `-take-last`, `-drop`, and `-drop-last`.

Furthermore, I ran into the above infloop issue while adding a similar `copy-sequence` to `-split-at`.

The solution: document that "dropping" operations like the above are only defined for proper lists, since it doesn't make sense to return the "last N elements" of an infinite list.

That would make the following test case in `dev/examples.el` invalid:

```el
(-map (-partial '-take 5) (-split-at 5 (-cycle '(1 2 3))))
```

This means the aforementioned Dash functions can return safely mutable tails using `copy-sequence`, and the argument against loss of generality is that these operations didn't make sense for infinite lists anyway.  How does that sound?

PR #355, opened 2020-12-31
Remove generated manual dash.info

basil-conto:
Both GNU ELPA and MELPA are able to and do generate `.info` manuals from corresponding `.texi` files.  This means that we no longer need to maintain `dash.info`, which is a hassle.

* `create-docs.sh`: Run under Bourne Shell.  Run batch Emacs with `-Q` flag.  Add line continuations to lines longer than 80 columns.  Don't generate `dash.info`.
* `dash.info`: Remove file.

Cc: @holomorph


magnars:
Excellent. Glad to be rid of it. 👍

tor. 31. des. 2020 kl. 17:13 skrev Basil L. Contovounesios <
notifications@github.com>:

PR #358, opened 2021-01-04
Various Texinfo manual improvements

basil-conto:
The main change here that I wanted to run by others first is changing the license of the Texinfo manual from [GPL](https://www.gnu.org/licenses/gpl-3.0.en.html) to [FDL](https://www.gnu.org/licenses/fdl-1.3.en.html).

The rest of the changes are summarised below.

Cc: @holomorph 

---

* `.gitignore`: Ignore locally compiled Info manual.

* `doc/fdl.texi`:
* `doc/gpl.texi`: New files containing GNU FDL and GPL in Texinfo format.

* `dash-template.texi`: Include their contents as appendices, and license manual under the GNU FDL.  Define placeholder flag `DASHVER` for the current version of the `dash.el` package.  Fix titles, subtitle, directory file entry, markup, and menus.  Improve wording and cross-references.  Reconcile manual contents with those of `readme-template.md`, particularly the change log and list of contributors.  Generate a function rather than concept index, which is not used.

* `dev/examples-to-info.el`: Set `DASHVER` in `dash.texi` to the current version of the `dash.el` package.

* `dash.texi`: Regenerate manual.

PR #359, opened 2021-01-04
Add Makefile

basil-conto:
I thought it would be nice to put all the scripting in a standard, central, powerful, and flexible place. WDYT?

**N.B.:** I've based this branch off #358, so I'm marking this PR as a draft until that one has landed.

---

* `Makefile`: New file subsuming `create-docs.sh` and `run-tests.sh`.
* `create-docs.sh`:
* `run-tests.sh`: Remove files.

* `.github/workflows/test.yml`:
* `dash-template.texi` (Contribute):
* `readme-template.md` (Contribute):
* `pre-commit.sh`: Refer to new Make targets.

* `dev/examples.el`: Add support for running byte-compiled tests in Emacs < 24.3, where `setf` was still defined in `cl-macs.el`.

* `README.md`:


basil-conto:
Now that #358 has landed and I've fixed byte-compiled tests on Emacs < 24.3, I'm marking this PR as ready to review.

Fuco1:
My only slight concerns are that while bash is fairly common on windows (e.g. coming with git) I'm not so sure about make.  Also what about Mac?

basil-conto:
> My only slight concerns are that while bash is fairly common on windows (e.g. coming with git) I'm not so sure about make.

It's definitely available on Windows, see e.g. the EZWinPorts created by Eli Zaretskii, one of the maintainers of Emacs: https://www.gnu.org/software/emacs/manual/html_node/efaq-w32/EZWinPorts.html.

I believe it's also available via WSL/MinGW/Chocolatey/etc., though I can't say how much harder any of these are to install compared to Bash.

Either way, my arguments for switching to Make are:
* We rely on CI to run tests anyway, since it's hard to test across all Emacs versions 24 through 28 (even I can't easily test on 24.1 and 24.2, and I'm on GNU/Linux).
* We rely on maintainers to regenerate docs anyway, since not everyone is able or bothered to install Texinfo.
* Contributors working on Windows able to do any of the above are likely also capable of installing Make.

> Also what about Mac?

Again, IIRC it's relatively straightforward to install via Homebrew, Xcode, or whatever is used these days.

Fuco1 (approved):

PR #362, opened 2021-01-05
Refactor/avoid recursion

perweij:
The recursive nature of `-partition-after-pred` causes problems when processing long lists where these limits are exceeded:
 `max-lisp-eval-depth`
 `max-specpdl-size`

I have tried to rewrite it as an iterative macro instead. It seems to work, and passes the tests, but if you have time, please check the code for possible improvements.


dash.el
```diff
@@ -1170,20 +1170,28 @@ returns the header value, but only after seeing at least one
 other value (the body)."
   (--partition-by-header (funcall fn it) list))
 
-(defun -partition-after-pred (pred list)
+(defmacro -partition-after-pred (pred list)
```
basil-conto, line 1173: It's not possible to turn a function into a macro with the same name, as that will break its existing callers. Also, Dash follows the naming convention where functions start with a single dash (e.g. `-foo`) and their macro counterparts start with two dashes (e.g. `--foo`).

I therefore suggest that, if you deem a macro is necessary, you provide that separately, and optionally reimplement the function in terms of the macro.

dash.el
```diff
@@ -1170,20 +1170,28 @@ returns the header value, but only after seeing at least one
...
-      (if (funcall pred (car list))
-          ;;split after (car list)
-          (cons (list (car list))
-                rest)
-
-        ;;don't split after (car list)
-        (cons (cons (car list)
-                    (car rest))
-              (cdr rest))))))
+  (declare (debug (pred pred)))
```
basil-conto, line 1175: AFAIK `pred` isn't a valid Edebug spec; see [`(info "(elisp) Instrumenting Macro Calls")`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Instrumenting-Macro-Calls.html) for info on `debug` declarations.

dash.el
```diff
@@ -1170,20 +1170,28 @@ returns the header value, but only after seeing at least one
...
+  (let ((r (make-symbol "result"))
+        (s (make-symbol "sublist"))
+        (l (make-symbol "list")))
+    `(let ((,l ,list))
+       (when ,l
+         (let* ((,r nil)
+		(,s nil))
```
basil-conto, line 1182: Please consistently indent with spaces, in accordance with the `indent-tabs-mode` setting in the project's `.dir-locals.el` file.

basil-conto (changes requested):
Thanks, it's always good to avoid recursion in Elisp if possible! :) But this still needs a bit of work. ;)

This probably counts as a non-trivial change (in terms of lines of code), so in order to install it we'd need you to assign the copyright to the Free Software Foundation, as Dash is copyrighted by the FSF.  Do you already have such an assignment on file?  If not, would you be willing to submit one?  The process is relatively simple and painless, and once it's done you'll be able to contribute to Emacs itself and other core packages as well. See [`(info "(emacs) Copyright Assignment")`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html) for more info on this.  You can find the relevant form here: https://git.savannah.gnu.org/cgit/gnulib.git/tree/doc/Copyright/request-assign.future.

TIA.

dash.el
Fuco1: The macro version should then be anaphoric in the predicate to follow the rest.

Personally these days I prefer to write the macros in terms of a function, the performance impact seems ignorable (until someone finds it slow enough).

basil-conto: > (until someone finds it slow enough)

;)

perweij:
Hi! Many thanks for checking this so quickly!

> If not, would you be willing to submit one? 

Done - I'll update here when I have received a reply.

dash.el
perweij: Hi! Thank you very much for the review! I'll rework it accordingly and be back,

dash.el
perweij: Good catch.

perweij: I removed it now, so should be OK.

dash.el
perweij: Now it is a function again, and a new anaphoric macro has been added:
  `(defun -partition-after-pred (pred list) ...)`
  `(defmacro --partition-after-pred (form list) ...)`

dash.el
perweij: Fixed, good catch.

perweij:
sorry @basil-conto I did not just deliberately request a new review - misunderstood the refresh icon, thought it would reload stuff but instead it sent you a request. Again, sorry.

basil-conto:
No worries! Now I know what it does too :)

basil-conto:
> Done - I'll update here when I have received a reply.

What's the status on this @perweij?  If you're waiting more than a week or two for a response from the copyright clerk, feel free to ping them.

basil-conto:
Ping.

perweij:
Hi! Now I have assigned the copyright to the Free Software Foundation. The process is completed, and I have received the final PDF file from FSF, so we should be able to proceed.  Sorry for the looong delay!

basil-conto (approved):

basil-conto:
> Sorry for the looong delay!

No worries, and thanks for the nice and efficient implementation!

PR #366, opened 2021-01-16
Avoid evaluating file-local variables

basil-conto:
Are there any objections to moving the file-local `eval`ed Font Lock keywords in `dev/examples.el` to `dash-fontify-mode`?  I think this is a bit cleaner and also avoids the "are you sure you trust these file-local variables" prompt every time you visit `dev/examples.el`.

* `dev/examples.el`: Move `font-lock-keywords` from `Local Variables`...
* `dash.el` (`dash--keywords`): ...to here, for `dash-fontify-mode`.


magnars:
👍

lør. 16. jan. 2021 kl. 15:32 skrev Basil L. Contovounesios <
notifications@github.com>:

PR #379, opened 2021-05-08
Rewrite -partition-after-pred to avoid recursion

akirak:
I encountered a situation where this function fails with the deep nesting error. I think it's generally better to avoid recursion whenever possible.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -1378,18 +1378,12 @@ other value (the body)."
 
 (defun -partition-after-pred (pred list)
   "Partition directly after each time PRED is true on an element of LIST."
-  (when list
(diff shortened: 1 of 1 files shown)


basil-conto:
> I think it's generally better to avoid recursion whenever possible.

Indeed, but have you tried the patch proposed in #362?

akirak:
I didn't know of that PR. Yes, I have worked on exactly the same issue. I don't have an anaphoric version, so the existing one would be better. Thanks.

basil-conto:
Thanks for confirming. Then let's wait a little bit longer for @perweij to respond, and in the meantime I'll check if their copyright papers are in order yet.

PR #380, opened 2021-05-27
Add/fix Edebug specifications.

phst:
Add specifications to macros that didn’t have them before.
Change some ‘form’ to ‘def-form’ specifications according to
https://www.gnu.org/software/emacs/manual/html_node/elisp/Specification-List.html.


basil-conto:
Thanks!

PR #381, opened 2021-06-08
Fix Edebug specification for -cut.

phst:
The first argument becomes a function name.  It can't be instrumented
because that would result in an invalid form such as

    ((edebug-after 0 1 FUNCTION) ...)


basil-conto:
Thanks!

PR #396, opened 2022-05-16
Symbols with destructuring expansions are not vars

### LINKED ISSUE #395: Spurious let-binding for &keys with empty plist
Cc: @Fuco1

Byte-compiling this example:

https://github.com/magnars/dash.el/blob/7fd71338dce041b352f84e7939f6966f4d379459/dev/examples.el#L1523

gives a warning as follows (from project root):

```console
$ emacs -Q -batch -L . -f batch-byte-compile dev/examples.el 2>&1 | grep \&keys
dev/examples.el:1421:14: Warning: Unused lexical variable `&keys'
```

Inspecting the macroexpansion of the offending form, we see:

```el
(let*
    ((#:--dash-source-326--
      (list 1 2 :bar 4 :foo 3))
     (a
      (pop #:--dash-source-326--))
     (b
      (pop #:--dash-source-326--))
     (&keys
      (car #:--dash-source-326--)))
  (list a b))
```

This seems to happen because `dash--match` considers all symbols as variables to be bound, even when `dash--match-cons` gives it something like a trailing `&keys` (i.e. with an empty plist).

basil-conto:
* `dash.el` (`dash--get-expand-function`): Return non-`nil` only if expansion symbol names an actual function.
(`dash--match-cons-1`): Simplify accordingly.
(`dash--match`): Don't bind symbols with expansion functions as if they were variables.

Fixes #395.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -2214,7 +2214,9 @@ matchers based on the type of the expression.
 Key-value stores are disambiguated by placing a token &plist,
 &alist or &hash as a first item in the MATCH-FORM."
   (cond
-   ((symbolp match-form)
+   ((and (symbolp match-form)
+         ;; Don't bind things like &keys as if they were vars (#395).
+         (not (functionp (dash--get-expand-function match-form))))
     (dash--match-symbol match-form source))
    ((consp match-form)
     (cond


Fuco1:
That's pretty unusual construct but if it gives warnings we should indeed fix it.

dash.el
```diff
@@ -2023,7 +2024,7 @@ SOURCE is a proper or improper list."
        ((cdr match-form)
         (cond
          ((and (symbolp (car match-form))
-               (functionp (dash--get-expand-function (car match-form))))
+               (dash--get-expand-function (car match-form)))
```
Fuco1, line 2027: I'm slightly undecided whether it's not better to call the `functionp` tests here (and other places this function is used), since the function is called `...get...` but here it serves as a predicate which might be a bit confusing.  And this matching code is quite complex, so maybe more verbosity is better.  WDYT?

Fuco1 (changes requested):

Fuco1 (approved):

Fuco1:
Perfect, I'm going to go ahead and merge it right now.  Thanks for spotting this!

basil-conto:
Thanks! Now to get around to the more pressing matters... :sweat_smile: 

> Thanks for spotting this!

Thank the tests you were nice enough to write ;).

PR #397, opened 2022-05-21
Fix clients of -compare-fn

basil-conto:
This improves the performance and correctness of clients of `-compare-fn`, but I'd like a review before merging because there are some slight changes in return values:

- `-contains?` now returns the matching tail of the list, just like `member`, instead of just `t`.
- `-union`, `-intersection`, and `-difference` now always return proper sets, without duplicates.
- ~~`-difference` continues to return multisets, with duplicates, because its docstring promises to return the same list it was given, modulo its intersection with another list.~~
- `-same-items?` now accepts multisets, i.e. lists with duplicates and/or different lengths, just like `seq-set-equal-p`.

Do these changes sound reasonable to you?

Should `-difference` also return a proper set, without duplicates?  I think its docstring can be interpreted either way, and so far it's been categorised as a set operation, so I don't think it's unreasonable for it to return a set.

Cc: @rejeep

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,12 +6,27 @@ See the end of the file for license conditions.
 
 ## Change log
 
-### From 2.19.1 to 2.19.2
+### From 2.19.1 to 2.20.0
 
 #### Fixes
 
 - Fixed a regression from `2.18` in `-take` that caused it to
   prematurely signal an error on improper lists (#393).
+- The functions `-union`, `-intersection`, and `-difference` now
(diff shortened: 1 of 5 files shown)


basil-conto:
> Should `-difference` also return a proper set, without duplicates? I think its docstring can be interpreted either way, and so far it's been categorised as a set operation, so I don't think it's unreasonable for it to return a set.

I've now gone with this, for convenience and consistency with other set operations.

I'll leave this PR open for comments/objections for another week or so before merging.

PR #398, opened 2022-05-27
Quote certain single quotes in docstrings

tarsius:
The byte-compiler recently got more fussy about quotes.

### DIFF
--- a/dash.el
+++ b/dash.el
@@ -2402,7 +2402,7 @@ patterns be optionally left out and derived from the key name in
 the following fashion:
 
 - a key :foo is converted into `foo' pattern,
-- a key 'bar is converted into `bar' pattern,
+- a key \\='bar is converted into `bar' pattern,
 - a key \"baz\" is converted into `baz' pattern.
 
 That is, the entire value under the key is bound to the derived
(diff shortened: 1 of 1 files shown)


Fuco1:
I loaded the functions and the docstrings look OK in Emacs 26, so this escaping is "backwards" compatible (no idea when this syntax was added or if it's new).  But since it works, I'm merging it.  Thanks!

tarsius:
It's not new and it was controversial when introduced (but I don't remember details).

basil-conto:
> It's not new

Indeed, `\=` has quoted the next character in docstrings since at least as far back as Emacs 23.  It's documented under [`(info "(elisp) Keys in Documentation")`](https://www.gnu.org/s/emacs/manual/html_node/elisp/Keys-in-Documentation.html).

> it was controversial when introduced (but I don't remember details)

Perhaps you are thinking of the automatic quote curving controlled by `text-quoting-style`, introduced in Emacs 25 (see [bug#20385](https://bugs.gnu.org/20385)?  Since then, it is necessary to quote backticks and apostrophes that should not be curved by `message`, `format-message`, et al.

tarsius:
> Perhaps you are thinking of the automatic quote curving controlled by text-quoting-style

Yes.

PR #291, opened 2019-03-15
Add `-to-head` and `-shuffle`

cireu:
Close https://github.com/magnars/dash.el/issues/283 and https://github.com/magnars/dash.el/issues/285

The tests won't pass in Emacs 24.2 or any lower version because we use `random` to create random seed. And this feature was introduced in Emacs 24.3. But I think the function of `-shuffle` can work well.

Issue #285, opened 2019-02-01
Add a function to take out an element from a list and return that element and the list


Fuco1:
Alternatively have a function to move an nth element to the front.

    (-to-front 4 '(0 1 2 3 4 5 6 7 8)) => '(4 0 1 2 3 5 6 7 8)

The user can easily recover the top and the rest by using `car` and `cdr`

PR #404, opened 2023-03-17
Add `-to-head` and `-shuffle` 

### LINKED ISSUE #283: [Feature Request]Add new function `-shuffle`

cireu:
Continue #291 

Close #283 & #285 

Cc: @debanjum

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,7 @@ Other list functions not fit to be classified elsewhere.
 * [`-butlast`](#-butlast-list) `(list)`
 * [`-sort`](#-sort-comparator-list) `(comparator list)`
 * [`-to-head`](#-to-head-n-list) `(n list)`
+* [`-shuffle`](#-shuffle-list) `(list)`
 * [`-list`](#-list-arg) `(arg)`
 * [`-fix`](#-fix-fn-list) `(fn list)`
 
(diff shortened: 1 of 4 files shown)


cireu:
I found that using the random function to set the seed may not always be effective, as it may be calling the srand or srandom function underneath, and the behavior of this function is platform-specific. In different processes, the same seed may not necessarily generate the same result.

So I plan to change the API of `-shuffle` to 

```
(fn LIST &optional RNG)
```

`RNG` should be a function that generates random numbers like the `random` provided by Emacs. In our test, we can use xorshift based RNG to create a deterministic random number sequence

basil-conto (changes requested):
Thanks for working on this!

> So I plan to change the API of `-shuffle` to
>
> ```
> (fn LIST &optional RNG)
> ```

If we really can't do without such an optional argument, then I at least don't think we should expose it as part of the advertised calling convention, at least not until it's really proven to be an essential part of the API.

Neither Clojure nor Racket, for example, expose such an argument in their standard shuffle routines.

So I'm not yet convinced that we need it.

> In our test, we can use xorshift based RNG to create a deterministic random number sequence

For better or worse, `dev/examples.el` is a mix of both user-facing examples and developer-facing unit/regression tests.

I believe that the user-facing examples should be as simple and transparent as possible, and need not be comprehensive.  Taking a custom RNG as argument for the sake of our test suite seems like the tail wagging the dog.  Unless someone brings forth a compelling argument for such an API that does not arise from testing needs, I think we can do without such an argument for now.

In our more comprehensive developer-oriented implementation tests, OTOH, we can do whatever we like.  We can have internal `ert-deftest`s at the end of `dev/examples.el`, `-shuffle` can call an internal function that takes an RNG argument for better testability, etc.  But the public API should not care about these details.

WDYT?

PR #411, opened 2023-12-05
add ->>as-> function for use with ->> 

zardoz03:
seems better and more consistent to use 
```lisp
(->> 1 (->>as-> i (+ i 1)))
```
rather than
```lisp
(->> 1 ((lambda (i) (+ i 1)))
```
\+ there's less consing of misc. closures to mimic the same functionality as `-as->`

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,7 @@ or readability.
 * [`->>`](#--x-optional-form-rest-more) `(x &optional form &rest more)`
 * [`-->`](#---x-rest-forms) `(x &rest forms)`
 * [`-as->`](#-as--value-variable-rest-forms) `(value variable &rest forms)`
+* [`->>as->`](#->>as->variable-rest-forms-value) `(value forms ... value)`
 * [`-some->`](#-some--x-optional-form-rest-more) `(x &optional form &rest more)`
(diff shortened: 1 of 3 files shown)


README.md
basil-conto, line 1: Note that this file is auto-generated from the docstrings in `dash.el` and the first three examples in `dev/examples.el`.

basil-conto (commented):
Thanks!

I won't be able to review this for a few days, so just a quick comment from me in the meantime.

README.md
zardoz03: So i'd have to rewrite the tests in order for it to work?
I didn't notice that when I was writing sorry.

Also I forgot to email FSF because other things cropped up -- shall have to ask about anonymous assignments + maybe PO boxes.

basil-conto: > So i'd have to rewrite the tests in order for it to work?

The first three tests should be self-contained and representative, then `make docs` regenerates both `README.md` and `dash.texi` accordingly.

If you have any difficulties or would rather avoid running `make docs` on your end, that's no problem, one of the maintainers or I can do so on the PR branch before merging.

> I didn't notice that when I was writing sorry.

No problem at all.

> Also I forgot to email FSF because other things cropped up -- shall have to ask about anonymous assignments + maybe PO boxes.

Good luck! Keep us posted.

PR #413, opened 2024-03-03
Create a Pcase pattern that works like `-let`.

### LINKED ISSUE #376: Make `dash--match` public?
Hello,

I am writing a macro that can optionally use Dash for destructuring, and have been using `dash--match` to create a list of bindings suitable for using in a `setq` or `let` form. This has worked well.

Please consider making this function public, so that it can be depended on by other packages.

Thank you.

okamsn:
This PR is the beginning of a `pcase` pattern that works like `-let`. For me, this would resolve #376 (which I opened).

Before I copy the tests for `-let` and work through the edge cases of the pattern, would you be willing to merge a `pcase` pattern that works like `-let`?

### DIFF
--- a/dash-pcase.el
+++ b/dash-pcase.el
@@ -1,0 +1,133 @@
+;;; dash-pcase.el --- A pcase pattern following `-let' -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2012-2024 Free Software Foundation, Inc.
+
+;; Author: Magnar Sveen <magnars@gmail.com>
+;; Keywords: extensions, lisp
+;; Homepage: https://github.com/magnars/dash.el
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


okamsn:
This is now ready for review.

I've copied the tests of `-let` for testing the pattern in `pcase-let` and `pcase`, except for a test of parallel bindings for `pcase`, which does not apply in that context.

Should the `&hash?` for only match hash table, or should it match if the value is `nil`?

Do you want any changes in how the code works or its style?

Thank you.

okamsn:
I have updated the code to use `dash--match-make-source-symbol` when `gensym` is not available.

How would you like to handle Emacs versions when `pcase-defmacro` is not defined? It was added in Emacs 25.1. I put the code in an `if` expression with a warning if `pcase-defmacro` isn't bound. Would you like it done some other way?

okamsn:
Another option would be making `dash-pcase.el` a separate package.

@basil-conto, do you have an opinion on how to handle `pcase-defmacro` not being defined in Emacs 24?

basil-conto:
> Another option would be making `dash-pcase.el` a separate package.

I don't really have an opinion on this; either way would work for me (I'm also quite unfamiliar with this part of Dash).

@Fuco1, @magnars any thoughts?

> @basil-conto, do you have an opinion on how to handle `pcase-defmacro` not being defined in Emacs 24?

IIUC this feature is about hooking Dash destructuring into `pcase`, when possible.

IMO if it is not possible, Dash does not need to do anything: users running Emacs 24 should neither expect Dash to invent a `pcase` feature that `pcase` does not itself provide; nor should they see warnings about Dash not being able to hook into `pcase`.

PR #414, opened 2024-05-10
Fix typo in --> docstring in dash.el

w08r:
Hola! 👋 

Not sure if interested in something so small, but was a tiny typo if you want a fix.


basil-conto:
Thanks!

PR #415, opened 2024-06-19
Added -keep-indexed-n.

mrtimdog:
Hi, I was after something I couldn't find, so I've added this:

`(defun -keep-indexed-n (num fn list)`

It's a bit like a `(-take n (-non-nil (-map-indexed fn list)))`, but won't evaluate `fn` unnecessarily.


basil-conto (changes requested):
Thanks for the contribution, and sorry for the long delay in responding.

> Hi, I was after something I couldn't find, so I've added this:
> 
> `(defun -keep-indexed-n (num fn list)`
> 
> It's a bit like a `(-take n (-non-nil (-map-indexed fn list)))`, but won't evaluate `fn` unnecessarily.

Could you please provide some motivation for adding this specific functionality? In other words, which use-cases did you encounter for such a function?

Dash provides a large number of functions that one might want to have combined in certain scenarios, but unless such combinations are needed often enough, they may not be worth adding separately.

For example, IIUC the behaviour of `-keep-indexed-n` can be replicated using `-each-while` to avoid superfluous evaluations of the `fn` argument (or `--keep`+`-take` if the extra evaluations are harmless).

Perhaps (as a compromise?) it would be more generally useful to add `-keep-indexed` first? At least there is precedent for such a function in [`clojure.core`](https://clojuredocs.org/clojure.core/keep-indexed). Does Clojure already have something like `-keep-indexed-n`?

BTW, do you have a copyright assignment to the FSF? If not, would you be interested in starting the process?

Issue #391, opened 2021-11-15
Bug: can't destructure to nothing


ComedyTomedy:
The following useless code probably shouldn't throw an error:

   (-setq (_) '(1))

Fails with "Attempt to set a constant symbol: nil".

That goes for anything where the destructuring form is entirely formed from () and _.

basil-conto:
Thanks for the report. Does PR #417 help?

ComedyTomedy:
I use underscores in leading arg position, like

    (-setq (_ _ c) '(1 2 3 4))

And `_` remains unbound

BUT there are inconsistent behaviours. After evaluating

    (-setq (_) '(1))

`_` is bound to nil, and after

    (-setq (_) '(1 2 3 4 5))

`_` is set to `2` (!).  If I try

    (-setq (a _ _) (number-sequence 1 5))

it sets `_` to `3`.

I'm assuming the ideal result is that any name that starts with `_` should never be bound. Basically all the happy paths work, actual variable names seem to be bound properly, but the `_` has a variety of results.

Might make sense to set up test cases that attempt to bind _a _b and _c etc in various positions and see what happens.

ComedyTomedy:
Thanks for working on this :) hope these little bugs aren't to discouraging, I think you've already improved on the previous behaviour.

basil-conto:
> I'm assuming the ideal result is that any name that starts with `_` should never be bound.

I would like to agree, except `(setq _ 1)` works that way, so `-setq` probably shouldn't deviate.

> Basically all the happy paths work, actual variable names seem to be bound properly, but the `_` has a variety of results.

Yes, I noticed some inconsistencies too, e.g. `(-setq _ 1)` vs `(-setq (_) '(1))`.

> Might make sense to set up test cases that attempt to bind _a _b and _c etc in various positions and see what happens.

You're right of course, but I'm unfamiliar with this part of Dash so deeper renovations might take (even more) time on my end.

basil-conto:
> Thanks for working on this :) hope these little bugs aren't to discouraging, I think you've already improved on the previous behaviour.

Not discouraging at all, I need all the help/data on this that I can get.
Thanks for the quick response!

PR #417, opened 2025-03-11
Fix destructuring (_)

basil-conto:
Currently `dash--match-cons` pops leading `_` placeholders before checking whether `match-form` contains only a single pattern.

This PR swaps these around, so that we check for a single pattern even if `match-form` is `(_)`.

This attempts to address #391.
Cc: @ComedyTomedy


basil-conto:
Hm, no, this isn't right: we still want `(_ _ x)` to count as a single pattern.

